PR #20, opened 2024-06-01
Fix r2modman/thunderstore profile importing on Linux

Rune580:
r2modman/thunderstore keep their profiles under config dirs (`~/.config/`) instead of data dirs (`~/.local/share/`).
this PR adds linux specific handling of this difference.

These changes fix r2modman/thunderstore profile importing on Linux


Kesomannen:
Thanks, I wasnt aware of this

I'll change the implementation to be more concise

PR #47, opened 2024-07-18
Migrate to Tauri v2 beta

Serial-ATA:
A lot of new Tauri apps use v2 nowadays, so I went ahead and migrated it if you're interested.

Everything seems to be working from my (not-so-extensive) testing, though I could've missed something. :)


Kesomannen:
Nice, I tried to do this before but everything exploded, will do some more testing then merge :)

PR #57, opened 2024-07-25
Import local mod from ZIP file

DaXcess:
This PR adds support for importing local mods from ZIP files directly, without having to extract them first.

This PR also immediately addresses a regression with importing local files, as the `open` tauri dialog returns an [`OpenDialogReturn<T>`](https://v2.tauri.app/reference/javascript/dialog/#opendialogreturnt) on files, instead of just a `string`/`string[]` (file paths).

The import strategy is actually pretty basic. Since most of the import logic is already in the "import from package" pipeline, importing from a zip simply just extracts the zip file into a temp directory, performs the "import from package" logic, and then deletes the extracted zip contents from the temp directory.

It would be a bit nicer if this actually read and extracted the files from the zip directly without relying on the "import from package" code, but that would require quite a bit of code duplication (at least for how the `installer.rs` is currently written, especially with the custom BepInExPack import code), hence why the simple approach for now.

### DIFF
--- a/src-tauri/src/manager/importer.rs
+++ b/src-tauri/src/manager/importer.rs
@@ -246,6 +246,12 @@ async fn import_local_mod(path: PathBuf, app: &AppHandle) -> Result<()> {
 
             local_mod.icon = plugin_path.join("icon.png").exists_or_none();
         }
+        LocalModKind::Zip => {
+            installer::install_from_zip(&path, &profile.path, &local_mod.name)
+                .context("failed to install local mod")?;
+
+            local_mod.icon = plugin_path.join("icon.png").exists_or_none();
+        }
         LocalModKind::Dll => {
             let file_name = path.file_name().unwrap();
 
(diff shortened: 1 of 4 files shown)


Kesomannen:
Nice! I think I'm going to merge the "from DLL" and "from zip" buttons as it already can detect which type the file is. The reason I split "from package" and "from DLL" was that it doesn't seem possible to create a file *or* directory select dialog.

Rewriting the mod installer to work without unzipping could work. In that case you would cache zipped versions of the mods instead of unzipping immediately after download. This would however slow down the cache install (which is the most used), as it would need to decompress the files before copying every time the mod is installed. Directly from the zip is definitely better for local mods though. Maybe a solution with generics and Fn traits is possible just to avoid the code duplication 🤔

DaXcess:
The "reading directly from zip" part was only targeted towards the local mod installation, so depending on how the application is written it shouldn't make the "normal" installs any different. I would have made the local imports work like that myself but I'm not familiar at all with the code base (and wanted to avoid code duplication), hence why I used the cache dir for temporarily storing the extracted bundle, and then deleting it (which I guess defeats the purpose of a cache, but it was the best place I could think of).

PR #75, opened 2024-08-13
See installed mods in the mod search list

### LINKED ISSUE #74: [Suggestion] mod downloaded indicator in mod browsing and uninstall button on condensed mods
r2 and thunderstore already have this, showing a checkmark at the rightside of the mod when already downloaded
image

it would be nice if something similar was added back instead of having to check the expanded section
image (26)

i would also like to suggest an unintall button on the condensed mods

example:
image (20)

DaXcess:
It can be annoying for some users to have to open the full mod description to see whether they have already installed a particular mod or not.

It looked like a bit of the work was already implemented for this feature, since the `isInstalled` prop was already being passed to the slot (although it was always set to `false`).

Was this not finished due to issues, or was this just a feature that was forgotten about?

This also adds a quick uninstall button, if the mod was already installed beforehand

This PR adds:
- Checking if a mod is installed by invoking `is_mod_installed`
- Adding a checkmark next to the mod name if it is installed
- Replacing the download button with an uninstall button if the mod is already installed
*(If the mod has dependencies, it warns the user before uninstalling, just like when uninstalling from the profile page)*

Maybe the ordering of the icons could be different, but that's up to personal preference.

Fixes #74

Example:
image

### DIFF
--- a/src-tauri/src/manager.rs
+++ b/src-tauri/src/manager.rs
@@ -23,8 +23,8 @@ use crate::{
     prefs::Prefs,
     thunderstore::{
         self,
-        models::FrontendProfileMod,
-        query::{self, IntoFrontendMod, QueryModsArgs, Queryable, SortBy, SortOrder},
+        models::{FrontendProfileMod, IntoFrontendMod},
+        query::{self, QueryModsArgs, Queryable, SortBy, SortOrder},
         BorrowedMod, ModRef, Thunderstore,
     },
     util::{
--- a/src-tauri/src/manager.rs
+++ b/src-tauri/src/manager.rs
@@ -24,7 +24,7 @@ use crate::{
     thunderstore::{
         self,
         models::FrontendProfileMod,
-        query::{self, QueryModsArgs, Queryable, SortBy, SortOrder},
+        query::{self, IntoFrontendMod, QueryModsArgs, Queryable, SortBy, SortOrder},
         BorrowedMod, ModRef, Thunderstore,
     },
     util::{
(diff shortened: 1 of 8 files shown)


Kesomannen:
I didn't fully add the checkmark thingy because of (perceived) performance reasons. Every time a mod is displayed it has to make a call to the rust side, which in turn has to wait for the mutex and do a linear search on a profile with possibly hundreds of mods. I didn't profile it in any way however, just kinda went with my gut feeling, so it might be completely fine.

I don't think the uninstall button belongs there - why would you want to uninstall something while your browsing mods? The PR was saying to add it in the profile list next to the enabled toggle, which I think is more reasonable.

DaXcess:
> I don't think the uninstall button belongs there - why would you want to uninstall something while your browsing mods? The PR was saying to add it in the profile list next to the enabled toggle, which I think is more reasonable.

I seem to have gotten confused by the term "condensed mods" referring to the same thing twice, but I now see that one of them referred to the mod search, and the other one to the profile modlist.

> I didn't fully add the checkmark thingy because of (perceived) performance reasons.

I thought about checking out the `query_profile`, `query_thunderstore` and `mod_query_result` commands/events, but I first decided to seek as to why this wasn't implemented in the first place. Maybe, if it's possible, these commands/events can already perform the check before returning the data to the frontend? It would reduce the amount of calls to the backend by a lot.
During my testing I haven't seen any lag whatsoever (might also be because of the async-ness of the calls), but having the backend do the checking beforehand is also cleaner and requires less code.

DaXcess:
36a97ef631e5578a80c524e939b5473c063289bf might be a better way to handle it, as it uses less hacky ways to insert the "is installed" state into the `FrontendMod` struct. Might still need some cleaning up, and it also still requires publicizing the "has_mod" function inside the `Profile` struct.

Kesomannen (commented):
Looks good. Publizing has_mod is fine, the modules are a bit of a mess anyway. I'm gonna remove the uninstall button since I plan on adding a right click menu to the mods (to open the same menu as the three dots) and having both feels redundant

PR #112, opened 2024-09-11
enable gzip for http requests

xiaoxiao921:
The speedup is quite substential. I went from downloading the Lethal Company mod index in about 1m+ to only 12s on my internet connection.


Kesomannen:
Oh wow, nice. Doesn't make a huge dent for me, but I don't see any negative to enabling it

PR #113, opened 2024-09-14
Fix paths rendering in right-to-left mode

notpeelz:
Fixes #100

image

[`U+200E`](https://graphemica.com/200E) is a left-to-right mark.
This preserves the "truncate from the left" behavior from right-to-left rendering.

Explanation: https://stackoverflow.com/a/66986273/1581233

Issue #100, opened 2024-08-31
Paths on Linux are always displayed with a trailing slash


notpeelz:
![image](https://github.com/user-attachments/assets/b25fa7fc-d04c-4375-b55d-1e327bea8383)

Expected: `/usr/bin/steam`
Actual: `usr/bin/steam/`

PR #116, opened 2024-09-14
Fix Path Text Rendering

TianMengLucky:
Add judgement, use ‘rtl’ when characters overflow, use ‘ltr’ when they don't.

Close #113 #100

### DIFF
--- a/src/lib/components/PathField.svelte
+++ b/src/lib/components/PathField.svelte
@@ -9,6 +9,7 @@
 	export let icon: string = 'mdi:folder';
 
 	$: hasValue = value && value.length > 0;
+	$: valueExcessive = value && value.length > 24;
 </script>
 
 <div class="flex items-center relative">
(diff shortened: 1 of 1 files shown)


src/lib/components/PathField.svelte
```diff
@@ -9,6 +9,7 @@
 	export let icon: string = 'mdi:folder';
 
 	$: hasValue = value && value.length > 0;
+	$: valueExcessive = value && value.length > 24;
```
notpeelz, line 12: That still doesn't get rid of the problem. RTL rendering affects "Other Neutral" Unicode characters.
It's best to keep `direction: rtl` and use the Right-to-Left Mark ([U+200E](https://graphemica.com/200E)) to suppress the bidi rendering.

Not sure why you opened another PR. Any changes that need to be done can be made in #100.

TianMengLucky: > That still doesn't get rid of the problem. RTL rendering affects "Other Neutral" Unicode characters. It's best to keep `direction: rtl` and use the Right-to-Left Mark ([U+200E](https://graphemica.com/200E)) to suppress the bidi rendering.
> 
> Not sure why you opened another PR. Any changes that need to be done can be made in #113.

Sorry, I was in a hurry.

PR #221, opened 2025-01-28
Fix Shimloader Installer's extraction of some DLLs

EssieFir:
Fixes some Shimloader files not being correctly extracted

Specifically:
`dwmapi.dll` which was overwritten by the version inside the "UE4SS" folder,
`UE4SS.dll` which wasn't extracted because the casing was different


Kesomannen:
👍

Issue #245, opened 2025-03-07
[FEATURE REQUEST] Use Steams `libraryfolders.vdf` to try to find install locations


AvarianKnight:
If users use library folders instead of the default `C:\Program Files (x86)\Steam\` folder they have to manually select the install location

Steam has an easily readable `.vdf` format (and there exists a [rust parsers](https://docs.rs/keyvalues-serde/latest/keyvalues_serde/) for it already), this could be used to get the `C:\Program Files (x86)\Steam\config\libraryfolders.vdf` and use that to automatically launch instead of having to manually override.

Spacemonkey164:
Beautiful idea! @AvarianKnight 

Surprisingly, I was just thinking about this today and was planning to post it myself. Thanks for sharing, upvoted! 👍🏼

PR #253, opened 2025-03-16
feat: automatically detect install location from libraryfolders.vdf

AvarianKnight:
This will automatically get the games installed folder from `libraryfolders.vdf` so the user doesn't need to manually set the install location.

this hasn't been tested on Linux, but works fine for me on Windows.

Closes https://github.com/Kesomannen/gale/issues/245


Kesomannen:
Nice. I cleaned up some of the code and also removed the steam library setting. I feel like it's redudant with these changes

Issue #226, opened 2025-02-02
Window isn't resizable on Linux


realestudios:
Currently running the latest KDE Plasma on EndeavourOS with X11, but I've also tried it on Wayland and Cinnamon X11 and had the same results. I downloaded Gale through the AUR.

The window can be maximized and snapped to the edges of the screen, but grabbing the edges of the window doesn't do anything.

Kesomannen:
Probably duplicate of #207

Test-Account666:
> Currently running the latest KDE Plasma on EndeavourOS with X11, but I've also tried it on Wayland and Cinnamon X11 and had the same results. I downloaded Gale through the AUR.
> 
> The window can be maximized and snapped to the edges of the screen, but grabbing the edges of the window doesn't do anything.

Just to confirm, if this is a duplicate, please try the modrinth app aswell

Apollonu:
Is there a solution to this issue? #207 does not exist

zikeji:
@Apollonu I got it to work by just calling the resize logic manually, but I have no idea if this has other implications for other Linux DE's, x11, etc.

https://github.com/zikeji/gale/commit/f100326b788ceab16cf3fde997cfd48b5ed39ad8

zikeji:
@Kesomannen Sorry to gravedig / remind of this closed issue - after frankly alot of troubleshooting by disabling and changing things one by one, I've found that removing the "unstable" feature and fixing the one rust reference requiring it resolves the issue.

https://github.com/zikeji/gale/commit/0fb9d715c0af0d726714116f164b2dc12a75e6d3

Of course, with no exposure to this repo beyond this, I have no idea if "unstable" was enabled for some other reason and removing it breaks something else.

PR #313, opened 2025-04-26
Remove unstable feature flag from Tauri

zikeji:
Refs #226 

The unstable feature flag breaks being able to resize the window on Linux. I found that the unstable flag was originally added way back when Tauri v2 was itself in beta but I couldn't find anything specifically requiring it nowadays.

https://github.com/Kesomannen/gale/pull/47
https://github.com/Kesomannen/gale/commit/777dc214ee8c3346668a919c1d09f3a7ca15f646

I did run the release CI against these changes with a few modifications, and will be daily driving this build and will update if I encounter any issues.

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -17,7 +17,7 @@ path = "src/main.rs"
 tauri-build = { version = "2", features = [] }
 
 [dependencies]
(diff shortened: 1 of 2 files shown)


Kesomannen:
Oh, nice

PR #359, opened 2025-06-20
Add PEAK support

Archibald-Wix:


Kesomannen:
Could you also bump the patch version number (to 1.8.3) so I can publish it right now? Change it in package.json, src-tauri/tauri.conf.json and src-tauri/Cargo.toml.

Archibald-Wix:
@Kesomannen bumped to 1.8.3 👍

PR #361, opened 2025-06-21
Better leasiblity for Game list

xammen:


Kesomannen:
I'm sorry but this isn't really the direction I would want to go with the UI

Issue #17, opened 2024-05-30
Scrolling down is easier than scrolling up


Test-Account666:
I *love* the latest update.

I noticed however that scrolling up is much harder than scrolling down.

In a modpack with 188 mods, scrolling down takes ~4 seconds, while scrolling back up takes ~30 seconds

Test-Account666:
This issue wouldn't be too big, if there was still a scroll bar present, but it was removed a few update back

Kesomannen:
This should be fixed now

Test-Account666:
Still not fixed. But the scroll bar is back :)

Damglador:
Yeah, can confirm that is not fixed

Damglador:
https://github.com/user-attachments/assets/c2983d31-4799-4e79-a3d9-502091d0612a

Snooziepup:
A "Scroll to top" button would be really beneficial

Flamy2000:
Seems to be working fine for me.

Kesomannen:
This is Linux-specific, might even be down to the desktop environment

Damglador:
> This is Linux-specific, might even be down to the desktop environment

I doubt that it's DE's fault. Wasn't able to find any similar issues on the internet, and didn't have it in other apps.

Kesomannen:
Is this fixed by now??

Damglador:
Nope

PR #378, opened 2025-07-08
fix: slow upwards scrolling

yobson1:
the cause was the constant viewport.scrollTo calls when attempting to scroll up

the jump prevention code is only going to do anything when the actual_height and expected_height aren't equal due to different height elements within the list, since they're all the same height they are always equal and d is always 0. the viewport.scrollTo calls weren't really doing anything other than causing #17. we can just check if d is not 0 before going ahead.

i presume this was a linux only issue due to the different embedded browsers used by tauri on different platforms, Edge WebView2 likely does literally nothing when you call viewport.scrollTo with the scrollTop of the element

before:
[Kooha-2025-07-08-14-02-07.webm](https://github.com/user-attachments/assets/b9907119-3314-4a64-b842-bd2dd39c4d30)

after:
[Kooha-2025-07-08-14-01-33.webm](https://github.com/user-attachments/assets/4b13ed18-40e2-4999-9f16-107b7665e14e)

### DIFF
--- a/src/lib/components/ui/VirtualList.svelte
+++ b/src/lib/components/ui/VirtualList.svelte
@@ -130,7 +130,9 @@
 			}
 
 			const d = actual_height - expected_height;
-			viewport.scrollTo(0, scrollTop + d);
+			if (d !== 0) {
+				viewport.scrollTo(0, scrollTop + d);
+			}
 		}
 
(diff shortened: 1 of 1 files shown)


Kesomannen:
Oh wow, finally. I hadn't looked very close at that component since I just copied it over from some place.

PR #395, opened 2025-07-28
Add PATAPON 1+2 REPLAY support

nyairobi:
[PATAPON 1+2 REPLAY](https://thunderstore.io/c/patapon-1-2-replay/) has just been added to thunderstore.io

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "gale",
-	"version": "1.9.2",
+	"version": "1.9.1",
 	"private": true,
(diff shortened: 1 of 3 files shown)


Kesomannen:
It's *fine* but scanning subfolders would be preferred (not even just for the purpose of mod managers but making general use more flexible). The track mode is susceptible to name collisions (not sure how you can guarantee that won't happen) and admittedly isn't that well-tested

Kesomannen (approved):

PR #436, opened 2025-09-13
Add BepisLoader support and Resonite game

hazre:
Adds support for [Resonite](https://store.steampowered.com/app/2519830) using the [BepisLoader](https://github.com/ResoniteModding/BepisLoader) mod loader. Resonite runs as two processes, the main engine and a separate renderer that talk to each other via IPC.

BepisLoader is basically a fork of BepInEx 6 that strips out all the Unity-specific stuff and works on .NET 9, which the main engine uses. The renderer side is handled by BepInExRenderer (just normal BepInEx 5).

The [RenderiteHook](https://github.com/ResoniteModding/RenderiteHook) package (needed for renderer mods) ships its own Doorstop and drops it in the Renderer folder, so BepInExRenderer can run without Gale needing to handle multiple loaders.

This approach is based on R2ModmanPlus's implementation (https://github.com/ebkr/r2modmanPlus/pull/1896, https://github.com/ebkr/r2modmanPlus/pull/1909).

For testing: Resonite is free on Steam, and if you want to check both BepisLoader and BepInExRenderer, try [AudioBridge](https://thunderstore.io/c/resonite/p/Knackrack615/AudioBridge/), it comes with both a Bepis plugin and a renderer plugin.

### DIFF
--- a/src-tauri/src/profile/install/installers/subdir.rs
+++ b/src-tauri/src/profile/install/installers/subdir.rs
@@ -161,12 +161,12 @@ impl<'a> SubdirInstaller<'a> {
         subdir: &Subdir,
         flatten: bool,
     ) -> Components<'p> {
-        // Only process if we're flattening and target starts with the trigger
+        // only process if we're flattening and target starts with the trigger
         if !flatten || !subdir.target.starts_with(subdir.name) {
             return components;
         }
 
-        // Get the suffix after the trigger name
+        // get the suffix after the trigger name
         let Some(suffix) = subdir.target.strip_prefix(subdir.name) else {
             return components;
         };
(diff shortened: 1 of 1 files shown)


Kesomannen:
Are you sure there isn't a way to make the installer so less code is duplicated? (extend the subdir installer instead?). I know the subdir code isn't the most readable, so if you want I could have a closer look myself and probably figure out a tighter solution

hazre:
Oh yeah my bad, I completely oversaw the extra_subdirs stuff and jumped the gun on `Subdir.target`. I changed back Subdir stuff back it's original types.

hazre:
Completely forgot about mod loader types in the front end, good catch!

Kesomannen:
Oops, I didn't mean to merge that like that. Oh well, should work fine anyway (you got some extra commits on your branch). I'll go ahead and merge this now!

hazre:
Thanks!

PR #439, opened 2025-09-16
Make custom launch arguments draggable

hazre:
depends on #438


Kesomannen:
Is there a particular reason you would need this? It feels like console arguments almost never care about order, and even then switching them manually doesn't take too much effort.

hazre:
The reason I added this was because I was watching my friend mentioning that the custom args wasn't working for them in gale and then told them about how the parameters and the values have to be on separate fields, then it was kinda painful to watch them basically have to remove everything and redo it. With dragging you can just add the values at the end and just edit the original parameter and drag it to the right location.

Frankly it's silly, so if you don't think it makes much sense, then feel free to close it.

PR #440, opened 2025-09-16
Add profile specific custom launch arguments

### LINKED ISSUE #408: [Feature Request] Profile-Specific Custom Launch Arguments
Would allow easy switching between profiles that do and do not use LocalMultiplayer, for example, which requires custom launch arguments to function.

hazre:
fixes #408 

depends on #439


Kesomannen:
I'm not sure this is the best direction when it comes to the rust side and commands. I would just stick it directly in the profile struct, sending it over with the other profile info, and have a command for editing the args. The prefs page and system is already quite messy so I don't think adding another, different way wouldn't be the biggest loss (ideally I would like to rewrite it completely, but it is working fine as is).

PR #628, opened 2026-06-18
Add mime type for .r2z files

### LINKED ISSUE #303: Assigned file types on Linux (Regression)
Right now, Gale for me is present as an app for opening ALL files. This is caused by `application/octet-stream`, if removed Gale is no longer in "Open with" menu for all files.

Another small issue is `x-scheme-handler/ror2mm`, it is not a default file type. I could implement it, but I have no clue how these files should look like.

Damglador:
- Install gale-mime.xml in /usr/share/mime/packages/
  - icon = gale
  - file extension = .r2z
- Replace mimeType application/zip with application/x-r2z (which is specified in gale-mime.xml

If package manager doesn't update the mime database automatically, one may run `sudo update-mime-database /usr/share/mime` to do that manually.

Closes https://github.com/Kesomannen/gale/issues/303

Now Gale will be the primary option for r2z files and won't show up on regular zip files at all.


Kesomannen:
Thanks for contributing! Just three things:
- Windows support would be nice. As it seems right now this would completely remove the handler on Windows?
- The `/static` directory is for frontend assets. Looking at the repo there aren't really a better place for it though, so I think this can be be placed in a new `assets` directory.
- Are you aware of how this works on AppImage and Flatpak?
  - If you want I can look into Flatpak myself, otherwise here's a relevant resource I found: https://docs.flatpak.org/en/latest/conventions.html#mime-files

Damglador:
From my understanding and from what Claude says (as it's unclear from the docs), Windows should just use what's defined in "ext", so it should be unchanged.

Flatpaks: https://docs.flatpak.org/en/latest/conventions.html#mime-files

For AppImages it probably depends on the manager used. Some may export MIME types, as documentation mentions:

>  How can I integrate AppImages with the system?
> Using the optional appimaged daemon, you can easily integrate AppImages with the system. The daemon puts AppImages into the menus, **registers MIME types**, icons, all on the fly. You can download it from this repository. But it is entirely optional.

Kesomannen:
> From my understanding and from what Claude says (as it's unclear from the docs), Windows should just use what's defined in "ext", so it should be unchanged.

Yes, but how does would it know what the mime type means? The xml file definitely looks Linux-only and doesn't even seem to be included in the installer unless Tauri is doing it implicitly

> Flatpaks: https://docs.flatpak.org/en/latest/conventions.html#mime-files

I saw your change, looks good but would be good to test it on a local build if you haven't already.

> For AppImages it probably depends on the manager used. Some may export MIME types, as documentation mentions:
>
> >  How can I integrate AppImages with the system?
> > Using the optional appimaged daemon, you can easily integrate AppImages with the system. The daemon puts AppImages into the menus, **registers MIME types**, icons, all on the fly. You can download it from this repository. But it is entirely optional

Alright, have you verified that the appimage actually contains the file though? (as it's only explicitly defined to be in the .deb)

Damglador:
Now it's everywhere. I can't test AppImage because `Error failed to bundle project 'failed to run linuxdeploy'` (even though I have the binary installed on the system), and without the AppImage I can't test flatpak (and because I don't want to install it), but `Gale.AppDir` does contain the mime type file now.

Damglador:
> Yes, but how does would it know what the mime type means? The xml file definitely looks Linux-only and doesn't even seem to be included in the installer unless Tauri is doing it implicitly

Windows doesn't use MIME types at all afaik, so yeah, it's Linux-only.

Kesomannen:
> Windows doesn't use MIME types at all afaik, so yeah, it's Linux-only.

Right, then the ext should be alright. I'm gonna try out the Flatpak build to see if it registers completely. If that works, I'll go ahead and merge!

Kesomannen:
Looks good! FYI the `Error failed to bundle project 'failed to run linuxdeploy'` error can be solved by settign `NO_STRIP=1`: https://github.com/tauri-apps/tauri/issues/14796

PR #637, opened 2026-07-05
Support Hexium

bwoebi:
This adds initial support for Hexium, valheim only.

Some code paths are hardcoding `valheim` to avoid any impact on other games.
This also tries centralizing all urls in backend.rs.

There's no support for dependents yet.


src-tauri/src/profile/import/mod.rs
bwoebi, line 1: Not sure what happened here? Was this CRLF before?

Kesomannen:
Will review today or tomorrow!

src-tauri/src/profile/import/mod.rs
Kesomannen: Probably, the codebase used CRLF before but I have been waging a war to remove it. However some vscode config on my Windows machine adds it back sometimes. I have tried to revert it to the best of my ability but seems like I missed it here.

I'll fix it on master

messages/en.json
```diff
@@ -21,6 +21,9 @@
 	"apiKeyPref_hasToken_has": "Click to override token",
 	"apiKeyPref_hasToken_no": "Not set",
 	"apiKeyPref_title": "Thunderstore API token",
+	"backendPref_title": "Download Servers",
```
Kesomannen: `Servers` shouldn't be capitalized here

messages/en.json
```diff
@@ -326,6 +329,9 @@
 	"navBar_link_modpack": "Export modpack",
 	"navBar_link_prefs": "Edit manager settings",
 	"navBar_link_profile": "Manage profile",
+	"otherServer_warn_title": "Download from other Server",
+	"otherServer_warn_content": "You are installing this mod from a source other than Thunderstore. This file has been scanned by the source hosting it, rather than by Thunderstore.",
+	"otherServer_warn_noremind": "Never warn about foreign installs",
```
Kesomannen: I feel like "Don't show again" is better to avoid repeating the sentiment of the dialog another time

src-tauri/src/profile/import/mod.rs
```diff
@@ -83,22 +85,29 @@ fn read_base64(base64: &str) -> Result<ImportData> {
...
-        .error_for_status()
-        .map_err(|err| match err.status() {
-            Some(status) if status == StatusCode::NOT_FOUND => {
-                eyre!("profile code is expired or invalid")
-            }
-            _ => err.into(),
-        })?
-        .text()
-        .await?;
+    let response = join_all(Backend::apply_all(
```
Kesomannen: Qualify this as future::join_all instead

src-tauri/src/profile/mod.rs
```diff
@@ -275,6 +286,17 @@ impl Profile {
         self.mods.iter().filter_map(ProfileMod::as_thunderstore)
     }
 
+    fn thunderstore_backend(&self) -> Backend {
```
Kesomannen: I feel like this could have a better name, as it feels weird that all mods in a profile doesn't necessarily come from the backend returned by `profile.thunderstore_backend()`. What about `has_hexium_mods` and return a bool instead?

src-tauri/src/thunderstore/backend.rs
```diff
@@ -0,0 +1,229 @@
...
+    pub fn get_owner_url(self, owner: &str, game: Game) -> String {
+        match self {
+            Backend::Thunderstore => {
+                format!("https://thunderstore.io/c/{}/p/{}/", game.slug, owner)
+            }
+            Backend::Hexium => format!("https://mods.valtools.org/teams/{}", owner),
+        }
+    }
+
+    pub fn get_mod_url(self, owner: &str, package: &str, game: Game) -> String {
```
Kesomannen: Would be good to use &PackageIdent here instead

src-tauri/src/thunderstore/backend.rs
```diff
@@ -0,0 +1,229 @@
...
+        match self {
+            Backend::Thunderstore => {
+                format!("https://thunderstore.io/c/{}/p/{}/", game.slug, owner)
+            }
+            Backend::Hexium => format!("https://mods.valtools.org/mods/1/{}/{}", owner, package),
+        }
+    }
+
+    pub fn get_download_url(self, owner: &str, name: &str, version: &str) -> String {
```
Kesomannen: .. and VersionIdent here

src-tauri/src/thunderstore/backend.rs
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    pub fn modpack_upload_baseurl(self) -> &'static str {
+        match self {
+            Backend::Thunderstore => "https://thunderstore.io/api/experimental",
+            Backend::Hexium => "https://mods.valtools.org/api/experimental",
+        }
+    }
+
+    pub fn apply_all<R>(f: impl Fn(Self) -> R, backends: Backends) -> Vec<R> {
```
Kesomannen: If it doesn't cause too much trouble with lifetimes this shouldn't be needed. Just have a `all()` method that returns an iterator over `Backend`, then consumers can use `.map` to achieve the same thing but without the intermediate `Vec`.

src-tauri/src/thunderstore/backend.rs
```diff
@@ -0,0 +1,229 @@
...
+        match backends {
+            Backends::All => vec![f(Backend::Thunderstore), f(Backend::Hexium)],
+            Backends::Thunderstore => vec![f(Backend::Thunderstore)],
+            Backends::Hexium => vec![f(Backend::Hexium)],
+        }
+    }
+}
+
+/// Registry of Thunderstore mods for the active game.
```
Kesomannen: The comment should communicate that this is for Thunderstore-"like" mods/sources and not just `thunderstore.io`.

src-tauri/src/thunderstore/cache.rs
```diff
@@ -129,9 +122,9 @@ pub fn write_packages(packages: &[&PackageListing], game: Game, prefs: &Prefs) -
     Ok(())
 }
 
-fn cache_path(game: Game, prefs: &Prefs) -> PathBuf {
+fn cache_path(game: Game, prefs: &Prefs, backend: Backend) -> PathBuf {
     prefs
         .data_dir
         .join(&*game.slug)
-        .join("thunderstore_cache.json")
+        .join(format!("{}_cache.json", backend))
```
Kesomannen: Can use `{backend}`

src-tauri/src/thunderstore/backend.rs
```diff
@@ -0,0 +1,229 @@
...
+                }
+            }
+        }
+    }
+
+    pub fn get_markdown_url(self, ident: &VersionIdent, cache: MarkdownKind) -> String {
+        match self {
+            Backend::Thunderstore => format!(
+                "https://thunderstore.io/api/experimental/package/{}/{}/{}/{}/",
+                ident.owner(),
```
Kesomannen, line 52: Can use `ident.path()` here instead

src-tauri/src/thunderstore/backend.rs
```diff
@@ -0,0 +1,229 @@
...
+                ident.name(),
+                ident.version(),
+                cache
+            ),
+            Backend::Hexium => format!(
+                "https://mods.valtools.org/api/experimental/package/{}/{}/{}/{}/",
```
Kesomannen, line 58: Same here

src-tauri/src/thunderstore/ident.rs
```diff
@@ -56,10 +56,6 @@ impl VersionIdent {
         (self.owner(), self.name(), self.version())
     }
 
-    pub fn path(&self) -> impl Display + '_ {
```
Kesomannen, line 59: Why was this removed?

src-tauri/src/thunderstore/mod.rs
```diff
@@ -78,13 +81,27 @@ impl<'a> From<(&'a PackageListing, &'a PackageVersion)> for BorrowedMod<'a> {
 pub struct ModId {
     pub package_uuid: Uuid,
     pub version_uuid: Uuid,
+    #[serde(
+        skip_serializing_if = "is_thunderstore_backend",
+        default = "thunderstore_backend"
```
Kesomannen: `Backend::default` already returns `Backend::Thunderstore` so we can use `serde(default)` without a helper function

src-tauri/src/thunderstore/mod.rs
```diff
@@ -78,13 +81,27 @@ impl<'a> From<(&'a PackageListing, &'a PackageVersion)> for BorrowedMod<'a> {
 pub struct ModId {
     pub package_uuid: Uuid,
     pub version_uuid: Uuid,
+    #[serde(
+        skip_serializing_if = "is_thunderstore_backend",
```
Kesomannen: It seems fine to serialize this even if it's the default value, would make it more explicit and potentially easier to migrate in the feature.

src-tauri/src/thunderstore/models.rs
```diff
@@ -9,13 +9,15 @@ use internment::Intern;
...
 
-use super::{PackageIdent, VersionIdent};
+use super::{Backend, PackageIdent, VersionIdent};
 use crate::{game::Game, profile::Profile};
 
 #[derive(Serialize, Deserialize, Debug, Clone, Eq)]
 pub struct PackageListing {
     #[serde(rename = "full_name")]
     pub ident: PackageIdent,
+    #[serde(skip_serializing, default = "thunderstore_backend")]
```
Kesomannen: Same comments here as on `ModId::backend`

src-tauri/src/thunderstore/backend.rs
```diff
@@ -0,0 +1,229 @@
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.write_str(match self {
+            Backend::Thunderstore => "thunderstore",
+            Backend::Hexium => "hexium",
+        })
+    }
+}
+
+impl Backend {
+    pub fn get_index_url(self, game: Game) -> Option<String> {
```
Kesomannen: Convention is to (usually) not prefix methods with `get_`. I think these would fit better without that, also matches existing `PackageListing` methods better

src/lib/components/mod-list/InstallModButton.svelte
```diff
@@ -18,26 +22,50 @@
...
 	let loading = $state(false);
+	let warnNoRemind = $state(false);
 
 	let disabled = $derived(mod.isInstalled || locked || loading);
 
-	let modId = $derived({
-		packageUuid: mod.uuid,
-		versionUuid: mod.versionUuid
-	});
+	let doInstall = async () => {
```
Kesomannen: (Unofficial) codebase convention is to use `async function` instead of the lambda syntax

src/lib/components/mod-list/ModDetails.svelte
```diff
@@ -123,8 +114,14 @@
 			</InfoBox>
 		{/if}
 
-		{#if mod.categories}
+		{#if mod.categories || mod.backend === Backend.Hexium}
 			<div class="mt-2 mb-1 flex flex-wrap gap-1">
+				{#if mod.backend === Backend.Hexium}
+					<div class="text-primary-200 rounded-full px-3 text-sm" style="background: #331b72; border: 1px solid #965dbe; padding-block: calc(var(--spacing) - 2px);">
```
Kesomannen: Please use tailwind classes for this. Background can be written as `bg-color-[#331b72]`

src/lib/components/mod-list/ModDetails.svelte
```diff
@@ -123,8 +114,14 @@
...
+						<img src="https://mods.valtools.org/favicon.ico" alt="" style="height: 1rem; display: inline;" />
```
Kesomannen: Same here, this can be written using tailwind

Kesomannen: Also we should bundle the icon ourselves (put in `/static`)

src-tauri/src/profile/export/mod.rs
```diff
@@ -48,6 +49,11 @@ pub struct R2Mod {
     #[serde(alias = "versionNumber")]
     pub version: R2Version,
     pub enabled: bool,
+    // This will make the field available for future comparison against the loaded mods.
+    // Defaults to Thunderstore for now.
+    #[allow(dead_code)]
+    #[serde(skip)]
+    pub backend: Backend,
```
Kesomannen, line 56: Not sure what the point of having it as a field if it's never used? Leaving the comment is fine but leaving the field feels weird. Commenting it out completely also skips the two annotation

src-tauri/src/profile/export/mod.rs
```diff
@@ -143,16 +150,14 @@ async fn export_code(app: &AppHandle) -> Result<Uuid> {
...
     };
 
     info!(len = base64.len(), "exporting profile code");
 
-    const URL: &str = "https://thunderstore.io/api/experimental/legacyprofile/create/";
-
     let response = app
         .http()
-        .post(URL)
+        .post(backend.profile_export())
```
Kesomannen, line 172: Not sure this the approach we want to go with. Currently this would make profiles with any hexium mods unable to be imported into other mod managers, even if the mods and versions are also available on thunderstore.

The ideal solution in my mind is to first check if each mod in the profile exists on thunderstore as well (and at the same version). If that's the case, uploading to thunderstore should take priority as with the changes below Gale will still check both, whereas r2 and tmm can resolve it normally.

In any case this shouldn't be done transparently, instead best would be if a warning is shown in the export dialog that the profile was uploaded to hexium and can only be imported with Gale. To do that you'd need to switch out the return type of the export command to provide more information to the frontend.

The same applies to modpacks. In that case a selection between thunderstore and hexium is probably necessary (and if some mods/versions are missing on either platform - either a warning that they won't be included or disabling the option completely), given the two require different API keys and such.

I'm sorry this wasn't caught by me earlier in the planning, the problem just occurred to me while reading the code.

src-tauri/src/thunderstore/backend.rs
bwoebi: It doesn't cause trouble and it's a smarter approach :-D

src-tauri/src/thunderstore/backend.rs
bwoebi: I find that obfuscates the actual url a bit (ident.path() hiding a slash in itself).

src-tauri/src/thunderstore/ident.rs
bwoebi: Because it's only usage was removed (see src-tauri/src/profile/install/queue.rs).

src-tauri/src/thunderstore/backend.rs
bwoebi: Do you prefer ident.path()? It just happens that in the thunderstore url and hexium url the package and name are consecutive, but it's not an inherent property here.

src/lib/components/mod-list/ModDetails.svelte
bwoebi: tailwind is confusing to me :-D

Anyway, I have `bg-[#331b72] border border-[#965dbe]` now, but I don't understand how I would express `calc(var(--spacing) - 2px)` in tailwind - I need to subtract a fixed amount of pixels from the general spacing to counteract the border size.

src-tauri/src/profile/export/mod.rs
bwoebi: It's not unused, it's used by the javascript code and then it'll satisfy some `{ name: string; backend: Backend }` types there.
It just happens to not have rust code behind - was the easiest way to fix it without adding a mapping from a type without backend to a type with backend

src-tauri/src/thunderstore/backend.rs
bwoebi: Agree, no idea why I did that.

src/lib/components/mod-list/InstallModButton.svelte
bwoebi: Right, functions in Javascript capture their scope too, not just lambdas, unlike rust.

src-tauri/src/thunderstore/backend.rs
bwoebi: Hmm, it still needs an intermediate Vec to return the list of backends, which might be 1 or 2.
Doing this doesn't change much in the end?

src-tauri/src/profile/export/mod.rs
bwoebi: Yeah, that makes sense. Will address in a separate commit.

bwoebi: No worries about not having caught it earlier, it didn't cause me to have to undo work :-D

So, yep, I just went with your ideal solution - checking whether any profile is fully supported by packages on thunderstore; if yes, I simply use thunderstore. If hexium was used, there's a note below the sharing code that it was used and the code can only be used by Gale.

For modpacks I default to thunderstore according to the same rules. Thunderstore is disabled as selection, if you use hexium-only mods. Otherwise you can still select hexium (for valheim mods), if that's what you want, but thunderstore is pre-selected.

The API keys are now separate too.

src-tauri/src/profile/export/commands.rs
```diff
@@ -49,16 +49,26 @@ pub fn export_file(dir: PathBuf, app: AppHandle) -> Result<()> {
...
     let mut manager = app.lock_manager();
 
     let game = manager.active_game;
     let profile = manager.active_profile_mut();
 
-    modpack::refresh_args(profile, game);
+    let hexium_exclusive = modpack::refresh_args(profile, &app, game);
 
-    Ok(profile.modpack.clone())
+    Ok(profile.modpack.as_ref().map(|args| ModpackInfo {
```
Kesomannen: Should be able to use `.cloned()` instead of `.as_ref()` to skip the clone inside of the closure.

src-tauri/src/profile/export/mod.rs
Kesomannen: Ah ok, that's fine then. A comment explaining that would be good

src-tauri/src/profile/mod.rs
```diff
@@ -275,6 +286,19 @@ impl Profile {
         self.mods.iter().filter_map(ProfileMod::as_thunderstore)
     }
 
+    /// Checks if any mods are hexium-exclusive mods
+    fn has_hexium_mods(&self, app: &AppHandle) -> bool {
+        self.thunderstore_mods().any(|(package, _)| {
+            let ident = &package.ident;
+            package.id.backend == Backend::Hexium
+                && app
+                    .lock_thunderstore()
```
Kesomannen: Locking inside a low-level profile method like this could easily lead to deadlocks as it's difficult to see it that it locks the thunderstore mutex internally. Also doing it for each element in the profile doesn't sound great performance-wise. I would change the method to take an already-locked `&Thunderstore` instead.

src-tauri/src/profile/mod.rs
```diff
@@ -275,6 +286,19 @@ impl Profile {
         self.mods.iter().filter_map(ProfileMod::as_thunderstore)
     }
 
+    /// Checks if any mods are hexium-exclusive mods
+    fn has_hexium_mods(&self, app: &AppHandle) -> bool {
```
Kesomannen: `has_hexium_mods` is a bit too general, `has_hexium_exclusive_mods` sounds better

src-tauri/src/profile/query.rs
```diff
@@ -153,6 +162,10 @@ impl Queryable for LocalMod {
         &self.name
     }
 
+    fn version(&self) -> semver::Version {
+        self.version.as_ref().map(|v| v.clone()).unwrap_or(semver::Version::new(0, 0, 0))
```
Kesomannen: This is a little cursed, how about `Queryable::version` returning an `Option` instead? Then, when comparing versions, order `None` below all `Some(version)`s

src-tauri/src/thunderstore/backend.rs
Kesomannen: > Hmm, it still needs an intermediate Vec to return the list of backends, which might be 1 or 2.

Can't you do something like `&[Backend::Thunderstore, Backend::Hexium].into_iterator()` or does that return a reference local variable? Defining the arrays as const and returning a `&'static` probably also works but is a bit cumbersome. Otherwise  `vec![Backend::Thunderstore, Backend::Hexium].into_iterator()` is also fine, at least avoids putting the function results in a `Vec` and needing consumers to call `into_iter()`.

> Doing this doesn't change much in the end?

I say it leads to more readable and idiomatic code as you can use the standard iterator methods without a mandatory "pseudo"-`map`.

src/lib/components/dialogs/ExportCodeDialog.svelte
```diff
@@ -53,6 +54,12 @@
 					onclick={copyCode}
 				/>
 			</div>
+
+			{#if code.backend !== Backend.Thunderstore}
+				<div>
+					{m.exportCodeDialog_galeExclusive()}
```
Kesomannen, line 61: Should be in an `<InfoBox type="warning">` or info

src-tauri/src/profile/export/commands.rs
bwoebi: No, that doesn't work: `cloned()` is defined on Option<&T>, but not on Option<T>. Best I can offer is `.as_ref().cloned()`, if that's what you want?

src/lib/components/dialogs/ExportCodeDialog.svelte
bwoebi: Thanks, that's exactly what I've been wondering about "how can that be styled" :-D Info looks great.

src-tauri/src/profile/query.rs
bwoebi: More manual work, but sure, can do that.
Oh, Option<T> has PartialOrd and orders None < Some anyway. Nice. :-D

src-tauri/src/thunderstore/backend.rs
bwoebi: Okay, done

src-tauri/src/profile/export/mod.rs
bwoebi: Yeah, right, forgot to mention "for UI"

src-tauri/src/profile/export/commands.rs
Kesomannen: Ah sorry, should be just `.clone()`

src-tauri/src/thunderstore/mod.rs
```diff
@@ -112,69 +118,108 @@ impl ModId {
...
     /// Whether packages have been succesfully fetched at least one since
     /// the last call to [`Thunderstore::switch_game`].
-    pub fn packages_fetched(&self) -> bool {
-        self.packages_fetched
+    pub fn packages_fetched(&self, app: &AppHandle, game: Game) -> bool {
+        let backends = app.lock_prefs().backends(game);
+        backends
+            .into_backend_slice()
+            .iter()
+            .map(|b| self.backend(*b).packages_fetched())
```
Kesomannen: Can just use .all() here without the map

src/lib/components/mod-list/InstallModButton.svelte
```diff
@@ -18,26 +22,50 @@
...
+	let modId = $derived({
+		packageUuid: mod.uuid,
+		versionUuid: mod.versionUuid,
+		backend: mod.backend
+	});
+
 	let versionsOpen = $state(false);
+	let dialogOpen = $state(false);
+	// svelte-ignore state_referenced_locally
+	let installPayload: ModId = $state(modId);
```
Kesomannen: `installPayload` is never used in a reactive way, it can be replaced with a parameter to doInstall instead.

src/lib/components/mod-list/InstallModButton.svelte
```diff
@@ -18,26 +22,50 @@
...
+		if (warnNoRemind) {
+			let prefs = await api.prefs.get();
+			prefs.backendSkipConfirm = true;
+			await api.prefs.set(prefs);
+		}
+		install(installPayload);
+		loading = true;
+		dialogOpen = false;
+	}
+	async function tryInstall(modId: ModId) {
```
Kesomannen: This logic also needs to be handled when the "quick" install button on the mod list items is used, as well as on Ctrl+click. Currently that's wired through browse/+page.svelte installLatest which always installs from Thunderstore (and has a type error currently). Cleanest solution is probably to move this all out to the browse page, with the confirm dialog.

Kesomannen (commented):
Looks good! Apart from the few very minor things, I'd like to see the quick install thing fixed but after that it seems good to merge

src/lib/components/mod-list/InstallModButton.svelte
bwoebi: I don't understand - doInstall still needs the parameter to be received from somewhere after the dialog?
I.e. install() happens, then if the dialog is needed, it needs to be stored intermittently, until the user confirms installing?

Or do you mean, you'd rather see it passed in `<Button color="accent" icon="mdi:download" onclick={doInstall}>`? i.e. make onclick take `() => doInstall(installPayload)`?

src/lib/components/mod-list/InstallModButton.svelte
bwoebi: Right, moved to this page. Thanks for pointing this out.

src/lib/components/mod-list/InstallModButton.svelte
Kesomannen: Ah I see, missed that. Although it still does not need the `$state` as it's only read inside the function.

src/lib/components/mod-list/InstallModButton.svelte
Kesomannen: :+1:

src/lib/components/mod-list/InstallModButton.svelte
bwoebi: These are the subtle differences between react and svelte I suppose, because in React it would be definitely needed :-)

PR #647, opened 2026-07-17
Display suggestions for mods

bwoebi:
Works exactly the same than dependencies.

image

Shows this on the mod page now.

### DIFF
--- a/src/lib/components/mod-list/ModDetails.svelte
+++ b/src/lib/components/mod-list/ModDetails.svelte
@@ -247,7 +247,7 @@
 	{/if}
 </Dialog>
 
-<Dialog title="Suggestions for {mod.name}" bind:open={suggestionsOpen}>
+<Dialog title="Suggested by {mod.name}" bind:open={suggestionsOpen}>
 	{#if mod.suggestions}
 		<ModCardList
 			mods={mod.suggestions.map((fullName) => ({ fullName, backend: mod.backend }))}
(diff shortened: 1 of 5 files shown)


src/lib/components/mod-list/ModDetails.svelte
```diff
@@ -226,6 +227,14 @@
 		)}
 	{/if}
 
+	{#if mod.suggestions !== null && mod.suggestions.length > 0}
+		{@render button(
+			'material-symbols:lightbulb-2',
```
Kesomannen: This should use the `mdi` icon set, not sure why dependencies doesn't already (there is a similar icon in material design icons: https://icon-sets.iconify.design/mdi/?icon-filter=dependency), I can fix that on master. You can use the `mdi:lightbulb` icon here.

src/lib/components/mod-list/ModDetails.svelte
```diff
@@ -238,6 +247,15 @@
 	{/if}
 </Dialog>
 
+<Dialog title="Suggested by {mod.name}" bind:open={suggestionsOpen}>
```
Kesomannen: This needs to be localized, again the dependencies dialog is wrong here.

Kesomannen (changes requested):

bwoebi:
Yeah, I copied this from Dependencies and considered these the gold standard :-D

Will change it.

Kesomannen:
That's a mistake on my end, not sure how that has slipped through the cracks lol

Kesomannen:
Fixed it on master

bwoebi:
Thanks, rebased and updated.

Kesomannen (approved):

Kesomannen:
:+1:

PR #648, opened 2026-07-17
Add support for multiple communities and add sfoy

bwoebi:
sfoy = Survival: Fountain of Youth

This is a Hexium-only community:
no confirmation dialogs happen there.

backends in the games.json default to ["Thunderstore"], overrides can now be done. Adding new communities is going to be trivial now.


src-tauri/src/game/mod.rs
bwoebi, line 1: Ugh, a CRLF case again :-/

src-tauri/src/profile/commands.rs
bwoebi, line 1: And here as well :-X

Kesomannen:
Running `cargo fmt` on master, that should fix all the CRLF. Might introduce some conflicts though

src-tauri/src/game/mod.rs
```diff
@@ -1,240 +1,255 @@
...
+    r2_dir_name: Option<&'a str>,
+
+    #[serde(borrow)]
+    mod_loader: ModLoader<'a>,
+
+    #[serde(borrow, default)]
+    platforms: Platforms<'a>,
+
+    #[serde(borrow, default = "thunderstore_backend_default")]
+    backends: Vec<&'a str>,
```
Kesomannen: Why not use `Backend` directly? That would make incorrect JSON an error instead of silently ignoring some backends in the conversion between JsonGame and GameData. Also, I think it's cleaner (and more inline with the other fields) to make this an `Option<Vec<_>>` and do the default logic in the conversion instead.

src-tauri/src/prefs/mod.rs
```diff
@@ -356,11 +356,15 @@ impl Prefs {
     }
 
     pub fn backends(&self, game: Game) -> Backends {
```
Kesomannen: I think this could use a better name, like:
```suggestion
    pub fn enabled_backends(&self, game: Game) -> Backends {
```

src-tauri/src/prefs/mod.rs
```diff
@@ -356,11 +356,15 @@ impl Prefs {
...
-        if game.slug == "valheim" {
-            self.game_prefs
-                .get(&*game.slug)
-                .map(|p| p.backend)
-                .unwrap_or_default()
+        if game.backends.contains(&Backend::Hexium) {
```
Kesomannen: If you implement `From<Backend> for Backends` you can write this neatly as:
```rust

match game.backends.as_slice() {
    [backend] => backend.into(),
    _ => self
        .game_prefs
        .get(&*game.slug)
        .map(|p| p.backend)
        .unwrap_or_default(),
}
```

src-tauri/src/thunderstore/backend.rs
```diff
@@ -27,21 +30,19 @@ impl Display for Backend {
...
-                if game.slug == "valheim" {
-                    Some(format!(
-                        "https://{}.hexium.gg/api/v1/package-listing-index/",
-                        game.slug,
-                    ))
-                } else {
-                    None
-                }
-            }
+        if game.backends.contains(&self) {
```
Kesomannen, line 30: If the only reason a `Game` needs to be passed is to check whether the `Backend` exists for it, I think it's better to let consumers handle checking that instead of coupling the behaviour of this with a whole `Game`.

src/lib/state/game.svelte.ts
```diff
@@ -48,18 +54,27 @@ class GamesState {
...
 
-		let backends = ['https://thunderstore.io/api'];
-		if (slug === 'valheim') {
+		let thunderstoreBackend = 'https://thunderstore.io/api';
+		let hexiumBackend = 'https://hexium.gg/api';
+		let backends: string[] = [];
+		if ((this.active?.backends?.length || 0) > 1) {
+			// for now assume this means both backends are available
 			let prefs = await api.prefs.get();
 			prefs.gamePrefs = new Map(Object.entries(prefs.gamePrefs));
```
Kesomannen, line 64: Seems like I passed on this before but what is going on here? I'm guessing the prefs instance is not being deserialized from the rust side with a `Map`? It looks pretty weird

src/lib/state/game.svelte.ts
```diff
@@ -48,18 +54,27 @@ class GamesState {
 			}
 		}
 
-		let backends = ['https://thunderstore.io/api'];
-		if (slug === 'valheim') {
+		let thunderstoreBackend = 'https://thunderstore.io/api';
```
Kesomannen: Rename these two to something like `thunderstoreUrl`, having `Backend` in the names despite not the type `Backend` (especially when other `Backend`s are floating around) is a bit confusing.

Also side note: I know this whole categories thing is really messy and now with the backend stuff repeats logic from the rust side. I'm actually planning on rewriting it in rust and using commands instead.

src/lib/state/game.svelte.ts
Kesomannen, line 1: `this.active?.backends?.length || 0` is used in quite a few places and looks a bit messy. I think it's worth to add a new field to this class like so:
```
activeBackends: Backend[] = $derived(this.active?.backends ?? []);
```
having `active == null` is not really a valid state anyway, so returning an empty backend list should be fine.

Kesomannen:
Surprised how few changes are in here, I guess I thought the Valheim implementation was a bit more hard coded. Good work, just one thing:

> This is a Hexium-only community:
> no confirmation dialogs happen there.

Not sure this is how I want it to be, it's still a non-Thunderstore download and someone who downloaded Gale from Thunderstore (or otherwise associates it with purely Thunderstore) might be "fooled" into believing Hexium-only games come from there as well. What do you think about showing a dialog when selecting the game for the first time? Wouldn't be too intrusive but still lets the user there is something they should know. We could also show the Hexium icon in the games list for these exclusive games.

src-tauri/src/game/mod.rs
bwoebi: I was thinking that we wouldn't want an error, when you'd ever drop something here.
If you prefer ossifying this, I can do that.

src-tauri/src/thunderstore/backend.rs
bwoebi: Not sure what you mean, `game.slug` is also accessed inside?

src/lib/state/game.svelte.ts
bwoebi: Yeah, that logic doesn't really nicely fit into frontend. But renaming to url for now.

bwoebi:
> thought the Valheim implementation was a bit more hard coded

Yeah, I tried to ensure to not make it too hard to rip it out later :-D

> What do you think about showing a dialog when selecting the game for the first time? Wouldn't be too intrusive but still lets the user there is something they should know. We could also show the Hexium icon in the games list for these exclusive games.

Let's show the hexium icon in the game selector, that should make it obvious enough, I'd say. With a title attribute explaining that it is from there and not thunderstore.
I don't think we need to make it that much more explicit? I can put a dialog there if you think it's needed.

src-tauri/src/prefs/mod.rs
bwoebi: Looks nice :-D Done.

src/lib/state/game.svelte.ts
bwoebi: Yes, it is weird.
It's copied from:
https://github.com/Kesomannen/gale/blob/3aa32514947c6c81785308982df2f8c9d1608bcf/src/routes/prefs/%2Bpage.svelte#L72-L76

You're actually passing a bare `{ [key: string]: GamePrefs }` from Rust, but declaring it as `gamePrefs: Map<string, GamePrefs>` in the `Prefs` type.

Does not really need to be a Map type, I'd say, just causes confusion.

src/lib/state/game.svelte.ts
bwoebi: good idea, done.

src-tauri/src/thunderstore/backend.rs
Kesomannen: Ah sorry, don't know how I didn't pay attention to that

src/lib/state/game.svelte.ts
Kesomannen: Ah ok, forgot that was done elsewhere as well. Leave that for now and I'll change it from a Map on master

messages/en.json
```diff
@@ -110,6 +110,7 @@
 	"gameSelect_content_1": "Your game missing? ",
 	"gameSelect_content_2": "Open an issue",
 	"gameSelect_content_3": " on our Github to let us know.",
+	"gameSelect_hexium_icon": "The community game only exists on hexium. There are no Thunderstore packages for it.",
```
Kesomannen: "The community game"?

How about:
```suggestion
This game only available on Hexium. All mods will be downloaded from there, rather than Thunderstore.
```

src-tauri/src/game/mod.rs
Kesomannen: It should be fine. The cached versions of the games list is used if there ever is a parse error in the live version, so outdated versions of Gale would continue to work if the backends were to change.

src-tauri/src/game/mod.rs
```diff
@@ -175,6 +176,13 @@ struct JsonGame<'a> {
 
     #[serde(borrow, default)]
     platforms: Platforms<'a>,
+
+    #[serde(default = "thunderstore_backend_default")]
+    backends: Vec<Backend>,
```
Kesomannen: Prefer making this an Option and doing the defaulting while converting instead

Kesomannen:
> Let's show the hexium icon in the game selector, that should make it obvious enough, I'd say. With a title attribute explaining that it is from there and not thunderstore.
I don't think we need to make it that much more explicit? I can put a dialog there if you think it's needed.

I'm fine with that for now. I'm planning on reworking the game selection menu anyway and will have to replace it then

messages/en.json
bwoebi: I reworded that sentence once too much :-D Fixing.

src-tauri/src/game/mod.rs
bwoebi: Like this, if I understood you correctly?

bwoebi:
Alright, adjusted the text now, should be good to go then.

src-tauri/src/game/mod.rs
Kesomannen: Yes, exactly

PR #649, opened 2026-07-17
Fix import/export of hexium mods

bwoebi:
We need to actually resolve the backend in code when reading the mods, otherwise it won't find hexium mods.

### DIFF
--- a/src-tauri/src/profile/export/mod.rs
+++ b/src-tauri/src/profile/export/mod.rs
@@ -52,9 +52,8 @@ pub struct R2Mod {
     pub enabled: bool,
     // This will make the field available to UI for future comparison against the loaded mods.
     // Defaults to Thunderstore for now.
-    #[allow(dead_code)]
-    #[serde(skip)]
-    pub backend: Backend,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub backend: Option<Backend>,
 }
 
 impl R2Mod {
@@ -113,7 +112,7 @@ pub(super) fn export_zip(profile: &Profile, writer: impl Write + Seek, game: Gam
                 ident,
                 version,
                 enabled,
-                backend: ts_mod.id.backend,
+                backend: None, // avoid serializing this
             }
         })
         .collect();
--- a/src-tauri/src/profile/export/mod.rs
+++ b/src-tauri/src/profile/export/mod.rs
@@ -50,10 +50,8 @@ pub struct R2Mod {
     #[serde(alias = "versionNumber")]
     pub version: R2Version,
     pub enabled: bool,
-    // This will make the field available to UI for future comparison against the loaded mods.
-    // Defaults to Thunderstore for now.
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub backend: Option<Backend>,
+    #[serde(default)]
+    pub source: Backend,
 }
 
 impl R2Mod {
@@ -112,7 +110,7 @@ pub(super) fn export_zip(profile: &Profile, writer: impl Write + Seek, game: Gam
                 ident,
                 version,
                 enabled,
-                backend: None, // avoid serializing this
+                source: ts_mod.id.backend,
             }
         })
         .collect();
--- a/src-tauri/src/profile/export/mod.rs
+++ b/src-tauri/src/profile/export/mod.rs
@@ -60,7 +60,11 @@ impl R2Mod {
     }
 
     pub fn into_install(&self, thunderstore: &Thunderstore) -> Result<ModInstall> {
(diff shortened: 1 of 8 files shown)


src-tauri/src/profile/import/mod.rs
```diff
@@ -52,36 +51,50 @@ pub struct ImportData {
...
         .map(BufReader::new)
         .context("failed to open profile manifest")?;
 
-    let manifest: ProfileManifest =
+    let mut manifest: ProfileManifest =
         serde_yaml::from_reader(reader).context("failed to read profile manifest")?;
 
+    // resolve backend now so that frontend knows where to fetch displaying from
+    for r2mod in manifest.mods.iter_mut() {
+        if let Ok(package) = thunderstore.find_mod(
```
Kesomannen: Can do `thunderstore.find_ident(&r2mod.version_ident)` here

src-tauri/src/profile/export/mod.rs
```diff
@@ -113,7 +112,7 @@ pub(super) fn export_zip(profile: &Profile, writer: impl Write + Seek, game: Gam
                 ident,
                 version,
                 enabled,
-                backend: ts_mod.id.backend,
+                backend: None, // avoid serializing this
```
Kesomannen: If I'm not mistaken, not serializing the backend field will cause Gale to "guess" the source of each mod (via the backfilling logic in `read_file`). I don't think this the behavior we want, if the exported profile has hexium mods that are also available on thunderstore, I would as a user expect users of my code to also get the hexium version of those mods. Gale already extends the r2x format with `game`, `ignored_version_updates` and `ignored_package_updates`, so adding an extra field doesn't matter in that way. I'll continue in `read_file`...

src-tauri/src/profile/export/mod.rs
```diff
@@ -52,9 +52,8 @@ pub struct R2Mod {
     pub enabled: bool,
     // This will make the field available to UI for future comparison against the loaded mods.
     // Defaults to Thunderstore for now.
-    #[allow(dead_code)]
-    #[serde(skip)]
-    pub backend: Backend,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub backend: Option<Backend>,
```
Kesomannen: I think `source` makes more sense as a field name, at least in the serialized format

src-tauri/src/profile/import/mod.rs
```diff
@@ -52,36 +51,50 @@ pub struct ImportData {
...
 
     let reader = File::open(temp_dir.path().join("export.r2x"))
         .map(BufReader::new)
         .context("failed to open profile manifest")?;
 
-    let manifest: ProfileManifest =
+    let mut manifest: ProfileManifest =
         serde_yaml::from_reader(reader).context("failed to read profile manifest")?;
 
+    // resolve backend now so that frontend knows where to fetch displaying from
```
Kesomannen: ... concerning the comments above in `R2Mod`. For every mod this should check: 1) if the mod has a `backend`, resolve it from there. If found keep the mod as is, otherwise check with any backend and override the field with that (exactly as it does currently).

The `backend` field also needs to respected in `R2Mod::into_install`

Kesomannen (changes requested):

src-tauri/src/profile/export/mod.rs
bwoebi: Yes. I also wasn't aware that adding this to the serialized struct was an option at all. In that case, adding the backend is certainly the superior choice. Will change.

src-tauri/src/profile/import/mod.rs
bwoebi: Ah, there's a method for that, nice :-D Only checked struct fields.

bwoebi:
Adjusted as requested, only does a lookup if not finding it in the primary backend.
This also now defaults to Thunderstore properly, i.e. perfect compatibility with R2ModMan profiles.

src-tauri/src/profile/import/mod.rs
```diff
@@ -52,36 +51,52 @@ pub struct ImportData {
...
```
Kesomannen: This comment is worded weirdly, I think a comment explaining the logic in the loop is better (like why there are two lookups and an `is_err()` involved is not obvious at first)

src-tauri/src/profile/import/mod.rs
Kesomannen: > The backend field also needs to respected in `R2Mod::into_install`

This still needs fixing

bwoebi: True, I missed that! Done now.

Kesomannen (approved):
Looks good!

PR #654, opened 2026-07-24
Fix touchpad not being able to open the info screen

NemGam:
## Issue

Mouse input:
IssueShowcase-mouse

Touchpad input:
IssueShowcase-touchpad

As can be seen from gifs above, clicking on a mod in the list via touchpad (laptop) immediately starts reordering elements instead of opening an info screen. 

This change introduces distance based constraint for drag and drop, fixing the issue.

## Fixed 
IssueShowcaseFixed

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
 	},
 	"type": "module",
 	"dependencies": {
+		"@dnd-kit/dom": "^0.4.0",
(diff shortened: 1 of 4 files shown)


Kesomannen:
Didn't know this was an issue, thanks for the fix!

PR #675, opened 2026-08-03
Prefer non-beta mod versions

bwoebi:
Specifically, semver versions with any prefix.

Meaning, when you have a mod with a beta version, installing said mod will install the latest released version.

Specific installed mods will indicate that there are beta updates, but "Update all" will never install the beta, unless the mod is currently beta.

You always can manually switch the version to any available version including betas.

### DIFF
--- a/src/lib/util.ts
+++ b/src/lib/util.ts
@@ -112,16 +112,22 @@ export function isOutdated(mod: Mod): boolean {
 	return mod.version !== mod.versions[0].name;
 }
 
+export function isNonReleaseVersion(version: string): boolean {
+	const metadataSeparator = version.indexOf('+');
+	const versionExtraSeparator = version.indexOf('-');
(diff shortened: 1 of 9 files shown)


src/lib/util.ts
```diff
@@ -112,6 +112,18 @@ export function isOutdated(mod: Mod): boolean {
...
+export function hasNonReleaseUpgrade(mod: Mod): boolean {
+	if (mod.versions.length === 0) {
+		return false;
+	}
+
+	if (mod.version?.includes('-')) {
+		return false;
+	}
+
+	return mod.versions[0].name.includes('-');
```
Kesomannen: Semver build metadata can contain hyphens: https://semver.org/#spec-item-10 so this needs to be more rigorous.

Kesomannen (changes requested):

src/lib/util.ts
bwoebi: Adjusted it to only check for hyphens before any + sign.

Kesomannen:
Looks good 👍

Kesomannen (approved):

PR #688, opened 2026-08-10
fix: persist UI state in tauri store instead of localStorage

### LINKED ISSUE #687: Bug: search input won't clear
clearing the search result makes it reappear on restart. Version v1.20.0

https://github.com/user-attachments/assets/b7057d68-072f-488b-96c8-b0d4782befc1

hazre:
fixes #687 by basically using tauri's store plugin instead of local storage. I copied persisted-state.svelte.ts from runed and edited it to work with tauri store instead to avoid changing the PersistedState consumers.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
 		"@sveltejs/kit": "^2.63.0",
 		"@sveltejs/vite-plugin-svelte": "^7.1.2",
 		"@tailwindcss/vite": "^4.3.0",
+		"@tauri-apps/cli": "^2.11.4",
 		"@types/node": "^25.9.2",
 		"prettier": "^3.8.1",
 		"prettier-plugin-svelte": "^3.4.1",
@@ -46,6 +47,7 @@
 		"@tauri-apps/plugin-os": "^2.3.2",
 		"@tauri-apps/plugin-process": "^2.3.1",
 		"@tauri-apps/plugin-shell": "^2.3.4",
+		"@tauri-apps/plugin-store": "^2.4.4",
 		"@tauri-apps/plugin-updater": "^2.10.0",
 		"bits-ui": "^2.15.4",
 		"clsx": "^2.1.1",
@@ -53,7 +55,6 @@
 		"js-convert-case": "^4.2.0",
 		"rehype-highlight": "^7.0.2",
 		"rehype-raw": "^7.0.0",
-		"runed": "^0.29.2",
 		"svelte-exmarkdown": "^5.0.2"
 	}
 }
--- a/src/lib/state/persisted-state.svelte.ts
+++ b/src/lib/state/persisted-state.svelte.ts
@@ -1,0 +1,100 @@
+import { LazyStore } from '@tauri-apps/plugin-store';
+
+const uiStore = new LazyStore('ui-state.json', { autoSave: true });
+
+type Serializer<T> = {
+	serialize: (value: T) => string;
+	deserialize: (value: string) => T | undefined;
+};
+
+type PersistedStateOptions<T> = {
+	/**
+	 * The serializer to use.
+	 *
+	 * @default { serialize: JSON.stringify, deserialize: JSON.parse }
+	 */
+	serializer?: Serializer<T>;
+};
+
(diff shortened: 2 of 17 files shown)


package.json
```diff
@@ -23,6 +23,7 @@
 		"@sveltejs/kit": "^2.63.0",
 		"@sveltejs/vite-plugin-svelte": "^7.1.2",
 		"@tailwindcss/vite": "^4.3.0",
+		"@tauri-apps/cli": "^2.11.4",
```
Kesomannen, line 26: Why was this added?

Kesomannen (changes requested):
Well done finding this issue, it has been bugging me for a while. For the fix, while this approach works, I would prefer to do something much simpler. Since we are only using the Tauri store and gale only has one window, a lot of the runed features like event synchronization, local and session storage go unused. When it's hoisted into gale like this it leads to a lot of dead and unnecessarily complex code.

I would like if you refactored it into a less configurable store just focused on the Tauri store. Also, that would allow for simplifying the async states, as I'm not a big fan of needing `waitForPersistedState()` (looks like the store just returns `null` before then?).

package.json
hazre: Because otherwise the cli has to be installed globally. Trying to run `pnpm tauri dev` says command doesn't exist. It's pretty much the default when u initialize a tauri project for the first time.

src/lib/theme.svelte.ts
```diff
@@ -311,10 +311,13 @@ export type Color =
 
 const root = document.querySelector(':root') as HTMLElement;
 export const colorFallbacks: Record<ColorCategory, Color> = {
-	accent: { type: 'system' },
```
hazre, line 314: I had to change this because it was freezing the UI on windows due to system not being available

hazre:
Okay I think this should be a lot more simpler now.

src/lib/theme.svelte.ts
Kesomannen: That's my bad, I had developed the system accent color on my Linux machine and didn't realize it was this broken on windows. I'll fix it on main so you can revert this here :)

package.json
Kesomannen: Ah ok, I misread it as the js bindings for the cli plugin which is only used on the rust side. Makes sense to add it here (although a bit outside of this PR's concern)

src/lib/state/persisted-state.svelte.ts
```diff
@@ -0,0 +1,102 @@
...
+	 */
+	serializer?: Serializer<T>;
+};
+
+const DEFAULT_SERIALIZER: Serializer<unknown> = {
+	serialize: JSON.stringify,
+	deserialize: JSON.parse
+};
+
+const instances = new Map<string, { hydrate(value: unknown): void }[]>();
```
Kesomannen: The only values being added to this are `PersistedState`, why not use that as the value type instead of any `{ hydrate(value: unknown): void }`?

src/lib/state/persisted-state.svelte.ts
```diff
@@ -0,0 +1,102 @@
...
+
+	get current(): T {
+		return this.#value;
+	}
+
+	set current(value: T) {
+		this.#value = value;
+	}
+
+	hydrate(value: unknown): void {
```
Kesomannen: I think making this take `string` and using `uiStore.entries<string>` is cleaner

src/lib/state/persisted-state.svelte.ts
```diff
@@ -0,0 +1,102 @@
...
+			this.#value = parsed;
+		}
+	}
+
+	#deserialize(value: string): T | undefined {
+		try {
+			return this.#serializer.deserialize(value);
+		} catch (error) {
+			console.error(`Error when parsing persisted store value for "${this.#key}"`, error);
+			return;
```
Kesomannen: `return` can be omitted here

hazre:
okay addressed all the issues

edit: woops forgot to revert the css theme thing

Kesomannen:
I'll fix the conflicts then merge it 👍

Kesomannen (approved):

PR #690, opened 2026-08-12
fix: svelte warnings

hazre:


src/lib/components/dialogs/LaunchOptionsDialog.svelte
```diff
@@ -15,7 +15,7 @@
 
 	let { open = $bindable(), options, gameName, onselect }: Props = $props();
 
-	let selectedOption = $state<string>(options[0]?.arguments ?? '');
```
hazre, line 18: This already gets set by the $effect below, so no need do it twice.

Kesomannen (approved):

Kesomannen:
Makes sense

PR #692, opened 2026-08-16
linux: lower the priority of steam.sh

Cloudef:
This breaks many linux systems that need to do runtime setup to actually run steam. Always prefer the `steam` in $PATH. If user uses flatpak steam that also should be preferred over `steam.sh`.

I'm not even sure if `steam.sh` should ever be used directly, but I'm going to leave it here as low prio.

Future improvement would be to be able to configure steam path on linux as well.

### DIFF
--- a/src-tauri/src/profile/launch/platform.rs
+++ b/src-tauri/src/profile/launch/platform.rs
@@ -105,18 +105,6 @@ fn create_base_steam_command() -> Result<Command> {
 
     // return Ok(Command::new("/home/keso/.local/share/Steam/steam.sh"));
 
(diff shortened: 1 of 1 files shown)


Redhawk18:
This would be very helpful to me :heart:

Kesomannen:
For some context, the main reason I added it was to match r2modman's behaviour, which to my knowledge always runs the script before trying anything else. I did have a quick look at the actual script but didn't realize it would break some distros.

I can look into why r2modman does use the script at all, and if there's any benefit to keeping it in Gale. But yeah, for now this seems very reasonable.

PR #710, opened 2026-08-27
Fix How to Fish path

marsn3:
Manually specifies the path for How to Fish as for some reason the actual game files are in a subdirectory in the main game directory.

(See SteamDB file listing for e.g. [Balatro](https://steamdb.info/depot/2379781/) and [How to Fish](https://steamdb.info/depot/4001891/) for reference, `How to Fish.exe` is listed under `How to Fish/How to Fish.exe` while Balatro's executable is listed by itself.)

### DIFF
--- a/src-tauri/games.json
+++ b/src-tauri/games.json
@@ -1419,7 +1419,8 @@
 		},
 		"platforms": {
 			"steam": {
-				"id": 4001890
+				"id": 4001890,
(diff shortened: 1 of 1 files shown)


Kesomannen:
`steam.dirName` is currently unused and should really be removed from the `game.json` file. What you want is to set `modLoader.fileTarget` to `./How to Fish`, I'll fix that on main

PR #720, opened 2026-09-10
Add cuphead support

JKLeckr:
## What is this?
This PR adds Cuphead support to Gale. Thunderstore's [ecosystem](https://github.com/thunderstore-io/ecosystem-schema/blob/master/games/data/cuphead.yml) supports it.

## Why is this useful?
It allows for folks to use Cuphead with Gale. Even if there is only one mod for Cuphead on Thunderstore currently as of writing, it opens the doors for easier Cuphead modding and more mods may appear.

## Did I test this?
I tested this with the CupheadArchipelago mod with the Steam version of Cuphead on Ubuntu 26.04. It works!

------

If there is anything missing that will prevent this PR from being accepted, feel free to reach out.


Kesomannen:
Thanks for the PR! Game additions should to use the `add_game.py` script which it seems like you have skipped, so there's a couple of things you'll have to fix up:
- Icons are no longer used for Thunderstore games. The icons are now instead fetched from thunderstore directly, and the only ones in this repository are for games that are only on Hexium.
- The formatting on the json looks a bit off, make sure the indenting is the same as elsewhere.
- `cuphead` should also be added to the list in `thunderstore.toml` so the gale thunderstore listing is properly updated.

JKLeckr:
Fixed. Evidently the json uses tabs instead of spaces and it looked correct in zed, but not on GitHub or vim where the tabs are rendered larger.
Double checked with the add_game.py and it seems to be consistent with what I have.
Removed the icon as well.

Kesomannen:
Looks good!

JKLeckr:
Yay! Just out of curiosity, when do you think the next release will touch down? It's so I can get an idea when to update my documentation for using a mod manager with Cuphead.

PR #728, opened 2026-09-12
Add local and remote dedicated server management. Tested with Valhost…

AzumattDev:
I added local and remote dedicated server management. Tested with Valhost and G-Portal servers for Valheim.

Gale can now launch a dedicated server installed on the same computer or deploy the active profile to a remote server. Dedicated server support only shows for games with a `dedicatedServer` entry in `src-tauri/games.json`, so this should scale pretty easily as more games are requested. I tried to keep the platform/executable detection close to what Gale already does.

The dedicated server dialog is opened from the arrow beside the profile launch button, similar to the existing modded/vanilla options.

For local servers, Gale finds the dedicated server through its configured platform, prepares the mod loader, and launches it with the active profile.

The server must already be installed. I didn't add anything to download it through Steam or other platforms. Closing Gale also doesn't close the server. I figured people would want control over that themselves.

Remote servers support:

* **SFTP (SSH)** with password or private-key authentication. The host-key fingerprint must be confirmed on first connection.
* **FTP / FTPS (automatic)** tries explicit FTPS first and only falls back to plain FTP if TLS isn't supported. Self-signed certificates require confirmation before credentials are sent.
* **Server directory** is whatever directory the host exposes. Use `/` if the account already opens at the server root.

They can use **Test connection** before deploying to make sure the connection/settings actually work.

**Deploy profile** first shows a preview of uploads, removals, unchanged files, and client-only mods that will be skipped.

Gale mirrors:

* `BepInEx/plugins`
* `BepInEx/patchers`
* `BepInEx/config`

Those directories are made to match the active profile. Missing files are uploaded, changed files are replaced, and files no longer in the profile are removed.

Disabled mods create `.old` files locally. Those aren't uploaded, and the enabled copy is removed from the server. Empty remote mod directories are cleaned up where possible.

Gale also skips generated BepInEx data/logs, Thunderstore package metadata (`README.md`, `CHANGELOG.md`, `icon.png`, `manifest.json`), and packages marked client-only by Thunderstore or Hexium.

If the host already manages BepInEx, Gale preserves its core/startup files. This is mainly for managed hosts where BepInEx is installed through the control panel or those files aren't writable.

Hosts that expose `plugins`, `patchers`, and `config` directly at the account root are detected automatically.

The preview uses the current remote files plus Gale's previous deployment manifest. If someone manually deletes a deployed file, Gale uploads it again even if the local profile hasn't changed.

Files outside the mirrored directories are only removed if Gale previously deployed them.

Cleanup failures don't fail an otherwise successful deployment. Gale keeps going and reports anything the host refused to delete. Connection/upload failures still stop deployment since the profile can't be considered synced.

Backend is under `src-tauri/src/profile/server`:

* `commands.rs` - Tauri commands, settings/secrets, progress events, and process state.
* `local.rs` - Finds and launches local dedicated servers.
* `deploy.rs` - Filters the profile, reads the remote tree, builds the preview, and applies it.
* `remote.rs` - Shared SFTP and FTP/FTPS operations.
* `manifest.rs` - `.gale-server-manifest.json` and relative-path validation.
* `runtime.rs` - Tracks the local server process.

Server settings are stored per profile. Passwords and private-key passphrases go through the OS credential store instead of the profile database.

Remote deployment uses BLAKE3 hashes and file sizes to avoid uploading unchanged files.

SFTP replacements use temporary/backup names and atomic rename where supported.

FTP uploads reconnect and retry transient transfer failures, but obviously don't have the same atomic replacement guarantees as SFTP.

The manifest tracks files from the last successful deployment. Normally it lives in the selected server directory. On restricted-root hosts it is stored under `config`.

Don't expand deletion beyond the three mirrored directories without keeping the manifest path validation and deployment preview protections.

Frontend pieces:

* `src/lib/api/profile/server.ts` - API bindings
* `src/lib/state/server.svelte.ts` - shared state
* `src/lib/components/dialogs` - dialogs
* `messages/en.json` - user-facing text through Paraglide

To add another supported game, define its dedicated-server platforms and default port in `src-tauri/games.json`.

Local launching also requires Gale to be able to find that platform's dedicated-server installation and executable.


AzumattDev:
Let me know if you need anything else from me.

docs/dedicated-servers.md
Kesomannen, line 1: While I'm not against documentation, there currently isn't any other docs like this in the codebase. The place to put it right now would be on the wiki. Making a new standard for in-codebase docs should be a separate addition to ensure coverage, not just on this feature.

src-tauri/src/profile/mod.rs
```diff
@@ -77,6 +78,7 @@ pub struct Profile {
     pub modpack: Option<ModpackArgs>,
     pub sync: Option<sync::SyncProfileData>,
     pub custom_args: String,
+    pub server_settings: server::config::DedicatedServerSettings,
```
Kesomannen, line 81: It feels wasteful to store this for every profile, even if it has never been launched as a server, or if the game doesn't support servers at all. Consider wrapping it in an Option (maybe even `Option<Box<...>>`?).

This would also allow you to remove the Default implementation on DedicatedServerSettings.

src-tauri/src/profile/launch/mod.rs
```diff
@@ -362,3 +372,7 @@ pub fn parse_steam_launch_options(steam_id: u32) -> Result<Vec<LaunchOption>> {
 
     Ok(launch_options)
 }
+
+pub(crate) fn apply_custom_args(command: &mut Command, args: &str) -> Result<()> {
```
Kesomannen, line 376: I would prefer making the module public instead of creating a wrapper

src-tauri/src/profile/launch/mod.rs
```diff
@@ -28,7 +31,7 @@ mod custom_args;
 #[cfg(target_os = "linux")]
 mod linux;
 mod mod_loader;
-mod platform;
+pub(crate) mod platform;
```
Kesomannen, line 34: I see you've used `pub(crate)` in multiple places, however this project is just a single crate with only one public export (all module definitions in `lib.rs` are private), so `pub(crate)` ends up effectively the same as `pub`. All the occurences should be changed to a bare `pub`.

src-tauri/src/profile/launch/mod.rs
```diff
@@ -226,6 +194,48 @@ impl ManagedGame {
...
         Ok(())
     }
+
+    #[cfg_attr(target_os = "windows", allow(unused_variables))]
+    pub(crate) fn apply_mod_loader_args(
+        &self,
+        command: &mut Command,
+        target_dir: &Path,
+        platform: Option<Platform>,
+        platforms: &Platforms<'_>,
```
Kesomannen, line 204: `self.game` already gives a reference to the platforms, this argument isn't needed

src-tauri/src/profile/server/commands.rs
Kesomannen, line 1: Ideally as much business logic as possible should be kept out of the `commands.rs` files. This file is definitely getting a bit too long and as too many responsibilities, try to push that towards the inner modules instead.

src-tauri/src/profile/server/commands.rs
```diff
@@ -0,0 +1,493 @@
...
+            process.profile_id,
+            process.game_slug,
+            process.server_dir,
+            child.clone(),
+        )?
+    };
+
+    emit_status(&app, &status);
+
+    notify_profile_changed(&app, process.profile_id);
```
Kesomannen, line 124: What changes to the profile data are we notifying for here? Also goes for the other usages in this file.

src-tauri/src/profile/server/runtime.rs
```diff
@@ -0,0 +1,155 @@
+use std::{
+    path::PathBuf,
+    process::Child,
+    sync::{Arc, Mutex},
+};
+
+use eyre::{Result, ensure};
+use serde::Serialize;
+
+pub type SharedChild = Arc<Mutex<Child>>;
```
Kesomannen, line 10: Using the async-friendly tokio Child would be better to avoid blocking.

src-tauri/src/profile/server/runtime.rs
```diff
@@ -0,0 +1,155 @@
...
+        let profile_id = running.profile_id;
+
+        let mut child = running
+            .child
+            .lock()
+            .expect("server child process mutex poisoned");
+
+        if child.try_wait()?.is_none() {
+            child.kill()?;
+            let _ = child.wait();
```
Kesomannen, line 128: Continuation on the comment at the top of this file -- the std's Child could block here and freeze the interface.

src-tauri/src/profile/server/commands.rs
```diff
@@ -0,0 +1,493 @@
...
+
+fn save_remote_settings(app: &AppHandle, settings: RemoteServerSettings) -> eyre::Result<()> {
+    let mut manager = app.lock_manager();
+    let profile = manager.active_profile_mut();
+    profile.server_settings.remote = settings;
+    profile.save(app, true)
+}
+
+fn spawn_process_watcher(app: AppHandle, child: SharedChild, pid: u32) {
+    thread::spawn(move || {
```
Kesomannen, line 409: SharedChild should be using tokio's async Child here which would allow this to be a standard Tokio task as opposed to a blocking std thread.

I would also like this funtion to be moved out of the commands file.

src-tauri/src/profile/server/config.rs
```diff
@@ -0,0 +1,144 @@
...
+    pub fn validate_local(&self, password: &str) -> Result<()> {
+        ensure!(
+            !self.server_name.trim().is_empty(),
+            "server name cannot be empty"
+        );
+
+        ensure!(!self.world.trim().is_empty(), "world name cannot be empty");
+        ensure!(self.port != 0, "server port cannot be 0");
+
+        if !password.is_empty() {
```
Kesomannen: Also, I don't quite understand why the password needs to validated. Isn't that entirely the server's responsibility? If you or the hosting platform set a password that doesn't pass the checks, why should Gale be the one to decide you're out of luck?

src-tauri/src/profile/server/local.rs
```diff
@@ -0,0 +1,146 @@
...
+    game::platform::Platform,
+    prefs::Prefs,
+    profile::{ManagedGame, launch, server::config::DedicatedServerSettings},
+};
+
+pub struct LocalServerProcess {
+    pub child: Child,
+    pub server_dir: PathBuf,
+    pub profile_id: i64,
+    pub game_slug: String,
```
Kesomannen, line 19: Since we can, I would prefer to store a reference to the full Game instance here.

src-tauri/src/profile/server/commands.rs
```diff
@@ -0,0 +1,493 @@
...
+        let prefs = app.lock_prefs();
+        let manager = app.lock_manager();
+
+        local::launch(manager.active_game(), &request.settings, &password, &prefs)?
+    };
+
+    let child: SharedChild = Arc::new(Mutex::new(process.child));
+
+    let pid = child
+        .lock()
```
Kesomannen, line 107: This can be done without locking if you just move it above the wrapping of `process.child`. Or, even better, just make `spawn_process_watcher` figure out the pid itself since it already has access to the `SharedChild`.

src-tauri/src/profile/server/commands.rs
```diff
@@ -0,0 +1,493 @@
...
+        .expect("server child process mutex poisoned")
```
Kesomannen, line 108: I appreciate the effort to be more descriptive with errors. However, codebase-standard currently is to just to `.unwrap` on Mutex-poisonings and I'd like to keep consistent. A repo-wide sweep to replace with `.expect` can be done later if needed.

src-tauri/src/profile/server/commands.rs
```diff
@@ -0,0 +1,493 @@
...
+pub async fn deploy_remote_server(
+    request: RemoteServerRequest,
+    app: AppHandle,
+) -> Result<DeploymentResult> {
+    ensure_no_pending_installs(&app)?;
+
+    let (profile_id, profile_dir, profile_mods) = {
+        let manager = app.lock_manager();
+        let profile = manager.active_profile();
+        (profile.id, profile.path.clone(), profile.mods.clone())
```
Kesomannen, line 164: `profile.mods.clone()` could be quite expensive and isn't required here, you can instead just move the client_only_mods inside the scope and return it. This also lets you use `Profile::thunderstore_mods` instead of filtering manually.

src-tauri/src/profile/server/commands.rs
```diff
@@ -0,0 +1,493 @@
...
+#[command]
+pub async fn preview_remote_server_deployment(
+    request: RemoteServerRequest,
+    app: AppHandle,
+) -> Result<DeploymentPreviewResult> {
+    ensure_no_pending_installs(&app)?;
+    let (profile_id, profile_dir, profile_mods) = {
+        let manager = app.lock_manager();
+        let profile = manager.active_profile();
+        (profile.id, profile.path.clone(), profile.mods.clone())
```
Kesomannen, line 204: Same `profile.mods.clone()` comment here

Kesomannen:
That's a big one 😅, I'll see if I have time to review today

src-tauri/src/profile/server/config.rs
```diff
@@ -0,0 +1,144 @@
...
+    pub server_name: String,
+    pub world: String,
+    pub port: u16,
+    pub public_server: bool,
+    pub crossplay: bool,
+    pub extra_args: String,
+    pub remote: RemoteServerSettings,
+}
+
+impl Default for DedicatedServerSettings {
```
Kesomannen, line 47: Default doesn't make much sense for this struct. I would much rather see it as an option on the profile struct and have the frontend construct an instance when it's actually needed (and with the context of the game, instead of always defaulting to "My Valheim Server").

src-tauri/src/profile/server/deploy.rs
```diff
@@ -0,0 +1,1308 @@
...
+    let mut cleanup_warnings = Vec::new();
+    for relative in &plan.removals {
+        let remote_path = layout.path(&settings.server_directory, relative);
+
+        match connection.remove_file(&remote_path) {
+            Ok(true) => removed_files += 1,
+            Ok(false) => {}
+            Err(error) => {
+                let warning = format!("Could not remove {relative}: {error}");
+                warn!("{warning}");
```
Kesomannen: tracing's macros support argument formatting directly:
```suggestion
warn!("Could not remove {relative}: {error}");
```

src-tauri/src/profile/server/manifest.rs
```diff
@@ -0,0 +1,67 @@
...
+    fn default() -> Self {
+        Self {
+            version: VERSION,
+            old_files_cleaned: false,
+            files: BTreeMap::new(),
+        }
+    }
+}
+
+pub fn is_safe_relative_path(path: &str) -> bool {
```
Kesomannen, line 33: `util::fs::is_enclosed` does the same thing but allows some more (safe) paths like `./file`

src-tauri/src/profile/server/secrets.rs
```diff
@@ -0,0 +1,58 @@
...
+
+fn entry(profile_id: i64, secret: ServerSecret) -> Result<Entry> {
+    Entry::new(
+        SERVICE,
+        &format!("profile-{profile_id}-{}", secret.suffix()),
+    )
+    .context("failed to access OS credential store")
+}
+
+pub fn set(profile_id: i64, secret: ServerSecret, value: &str) -> Result<()> {
```
Kesomannen, line 33: These functions could be methods on the `ServerSecret` struct.

src/lib/api/profile/server.ts
```diff
@@ -0,0 +1,137 @@
+import { invoke } from '$lib/invoke';
+
+export type ServerLocation = 'local' | 'remote';
```
Kesomannen, line 3: Type declarations should be moved to the `types.ts` file

src/lib/api/profile/server.ts
```diff
@@ -0,0 +1,137 @@
...
+	| { state: 'stopped' }
+	| {
+			state: 'running';
+			profileId: number;
+			gameSlug: string;
+			pid: number;
+			serverDir: string;
+	  };
+
+export function getSettings() {
```
Kesomannen, line 79: These should match the other files under `api/`, that is use the arrow syntax.

src/lib/components/dialogs/DedicatedServerDialog.svelte
```diff
@@ -0,0 +1,558 @@
...
+
+	const DEFAULT_SFTP_PORT = '22';
+	const DEFAULT_FTP_PORT = '21';
+	const FALLBACK_SERVER_PORT = 2456;
+	const MAX_PORT = 65535;
+
+	type Props = { open?: boolean };
+	let { open = $bindable(false) }: Props = $props();
+
+	let location = $state<ServerLocation>('local');
```
Kesomannen, line 41: Instead we should be able to use a `$state<DedicatedServerSettings>` and keep the reactivity instead of defining every single field as a separate variable. If needed, you can use Svelte's function bindings to map a value, for example for `port`: https://svelte.dev/docs/svelte/bind#Function-bindings. This also avoids the conversion methods like `settings`, `remoteSettings` and the bulk of the logic in `loadSettings`.

src/lib/components/dialogs/DedicatedServerDialog.svelte
```diff
@@ -0,0 +1,558 @@
...
+		return result;
+	}
+
+	async function save() {
+		const current = await checkedSettings();
+		if (!current) return;
+		saving = true;
+		try {
+			await api.profile.server.setSettings(current);
+			await message(m.dedicatedServerDialog_saved());
```
Kesomannen, line 240: This should use a toast rather than an OS popup. Also goes for the other "success messages" in this file.

src/lib/components/dialogs/DedicatedServerDialog.svelte
```diff
@@ -0,0 +1,558 @@
...
+				});
+				return;
+			}
+			if (!(await deploymentPreviewDialog.openFor(preview))) return;
+
+			const result = await requestWithTrust(current.remote, () =>
+				api.profile.server.deployRemote(current.remote, remotePassword, rememberRemotePassword)
+			);
+			if (!result) return;
+			await api.profile.server.setSettings(current);
```
Kesomannen, line 315: Looks like `api.profile.server.deployRemote` already saves the settings, so this is unneeded.

src/lib/components/dialogs/DedicatedServerDialog.svelte
```diff
@@ -0,0 +1,558 @@
...
+	}
+</script>
+
+<Dialog title={m.dedicatedServerDialog_title()} bind:open large>
+	<p class="text-primary-600 dark:text-primary-300 mt-1">
+		{m.dedicatedServerDialog_content()}
+	</p>
+
+	{#if loadingSettings}
+		<div class="text-primary-500 mt-5">{m.dedicatedServerDialog_loading()}</div>
```
Kesomannen, line 330: Can use a spinner here :)

src/lib/components/dialogs/DedicatedServerDialog.svelte
```diff
@@ -0,0 +1,558 @@
...
+		<TabsMenu
+			bind:value={location}
+			options={[
+				{ value: 'local', label: m.dedicatedServerDialog_locationLocal() },
+				{ value: 'remote', label: m.dedicatedServerDialog_locationRemote() }
+			]}
+		>
+			<Tabs.Content value="local">
+				<div class="mt-4 flex flex-col gap-3">
+					<div>
```
Kesomannen, line 341: Having the Label above the field is not consistent with the rest of the UI. Either use modpack-page style `FormField`s or use the horizontal `Label` layout used elsewhere.

src/lib/components/dialogs/DedicatedServerDialog.svelte
```diff
@@ -0,0 +1,558 @@
...
+								<span class="truncate">
+									{deploymentProgress.operation === 'upload'
+										? m.dedicatedServerDialog_progressUpload({ path: deploymentProgress.path })
+										: m.dedicatedServerDialog_progressRemove({ path: deploymentProgress.path })}
+								</span>
+								<span class="shrink-0"
+									>{deploymentProgress.completed}/{deploymentProgress.total}</span
+								>
+							</div>
+							<Progress.Root
```
Kesomannen, line 508: Since a progress bar is now used in both InstallPopover and here, it makes sense to extract it to a shared component.

src/lib/components/dialogs/DeploymentStats.svelte
```diff
@@ -0,0 +1,43 @@
...
+			value /= KIBIBYTE;
+			unit++;
+		}
+		const decimals = value >= 10 ? 1 : 2;
+		return `${value.toFixed(decimals)} ${units[unit]}`;
+	}
+</script>
+
+<div class="mt-5 grid grid-cols-3 gap-3">
+	{#each [{ icon: 'mdi:cloud-upload', label: m.deploymentStats_upload(), value: uploaded, detail: formatBytes(bytes) }, { icon: 'mdi:trash-can-outline', label: m.deploymentStats_remove(), value: removed, detail: m.deploymentStats_serverFiles() }, { icon: 'mdi:check-circle-outline', label: m.deploymentStats_unchanged(), value: unchanged, detail: m.deploymentStats_noTransfer() }] as stat}
```
Kesomannen, line 31: Make this a field in the script section instead of inlining it

src/lib/components/dialogs/DeploymentStats.svelte
```diff
@@ -0,0 +1,43 @@
...
+	type Props = {
+		uploaded: number;
+		bytes: number;
+		removed: number;
+		unchanged: number;
+	};
+
+	let { uploaded, bytes, removed, unchanged }: Props = $props();
+
+	function formatBytes(bytes: number) {
```
Kesomannen, line 16: Use shortenFileSize from util.ts instead

src-tauri/src/profile/server/config.rs
```diff
@@ -0,0 +1,144 @@
...
+pub enum RemoteProtocol {
+    #[default]
+    Sftp,
+    Ftp,
+    Ftps,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(default, rename_all = "camelCase")]
+pub struct DedicatedServerSettings {
```
Kesomannen, line 36: As I wrote in the comment in `commands.rs`, I think the structure of these settings have to be redesigned. I have these two things in mind:

- `server_name`, `world`, `public_server`, `crossplay` and `password` (the local password that is) seem Valheim-specific. A better cross-game solution is needed here.
- Having `remote` and `location` as a member of `DedicatedServerSettings` doesn't make much sense. When `location` is `ServerLocation::Local`, `RemoteServerSettings` isn't used, and when `location` is `ServerLocation::Remote`, none of the other fields on this struct are. This can be better encoded with the type system. How about properly splitting the two and moving the location field? Something like this:

```rust
// fields here still need to be made cross-game
pub struct LocalServerSettings {
    pub server_name: String,
    pub world: String,
    pub port: u16,
    pub public_server: bool,
    pub crossplay: bool,
    pub extra_args: String,
}

pub struct RemoteServerSettings {
    pub protocol: RemoteProtocol,
    pub host: String,
    pub port: u16,
    pub username: String,
    pub server_directory: String,
    pub authentication: RemoteAuthentication,
    pub private_key_path: String,
    pub trusted_host_key: Option<String>,
    pub trusted_invalid_certificate_host: Option<String>,
}

// this would be stored by the profile
pub struct ProfileServerSettings {
    pub location: ServerLocation;
    pub local: LocalServerSettings;
    pub remote: RemoteServerSettings;
}
```
One could also make the `ProfileServerSettings` an enum for more correctness, however then the settings aren't remembered when you switch back and forth which is a pretty important QoL.

src/lib/components/dialogs/DeploymentStats.svelte
```diff
@@ -0,0 +1,43 @@
...
+	{#each [{ icon: 'mdi:cloud-upload', label: m.deploymentStats_upload(), value: uploaded, detail: formatBytes(bytes) }, { icon: 'mdi:trash-can-outline', label: m.deploymentStats_remove(), value: removed, detail: m.deploymentStats_serverFiles() }, { icon: 'mdi:check-circle-outline', label: m.deploymentStats_unchanged(), value: unchanged, detail: m.deploymentStats_noTransfer() }] as stat}
+		<div class="bg-primary-100 dark:bg-primary-900 rounded-lg p-3">
+			<div class="text-primary-500 dark:text-primary-400 flex items-center gap-2 text-sm">
+				<Icon icon={stat.icon} />
+				{stat.label}
+			</div>
+			<div class="text-primary-800 dark:text-primary-200 mt-1 text-xl font-semibold">
+				{stat.value}
+			</div>
+			<div class="text-primary-500 text-sm">{stat.detail}</div>
```
Kesomannen, line 40: Contrast here is slightly too low

src/lib/components/dialogs/DeploymentPreviewDialog.svelte
```diff
@@ -0,0 +1,108 @@
...
+			<summary class="text-primary-700 dark:text-primary-300 cursor-pointer font-medium">
+				{m.deploymentPreview_fileChanges({
+					count: preview.uploadFiles.length + preview.removeFiles.length
+				})}
+			</summary>
+			<div
+				class="border-primary-300 dark:border-primary-600 bg-primary-50 dark:bg-primary-900 mt-2 max-h-64 overflow-auto rounded-lg border p-3 font-mono text-sm"
+			>
+				{#if preview.uploadFiles.length > 0}
+					<div class="text-primary-500 mb-1 font-sans font-semibold">
```
Kesomannen, line 77: Contrast a bit low

src/lib/components/dialogs/DeploymentPreviewDialog.svelte
```diff
@@ -0,0 +1,108 @@
...
+		{/if}
+
+		<details class="mt-4">
+			<summary class="text-primary-700 dark:text-primary-300 cursor-pointer font-medium">
+				{m.deploymentPreview_fileChanges({
+					count: preview.uploadFiles.length + preview.removeFiles.length
+				})}
+			</summary>
+			<div
+				class="border-primary-300 dark:border-primary-600 bg-primary-50 dark:bg-primary-900 mt-2 max-h-64 overflow-auto rounded-lg border p-3 font-mono text-sm"
```
Kesomannen, line 74: Could use slightly padding to better match style of the markdown code blocks. Border should also be skipped.

dannzen:
just as a side note, it really depends on the server setup.

iam using currently ChatGPT to manage my gale "synced" server... there are configs which are user depended inside the profile, others are server depended and would overwrite any server changes inside the config. Some configs are created on first load or like other mods, need a biome reset.

normally i would say, the best way to give that issue to the docker dedicated valheim team and let them implement a gale sync, where you only have to enter the profile id. I wrote 2 scripts as docker watchers, which are pooling the gale id and are autoupdating the server if no one is online. Otherwise are pushing a server annoucment for a planned reboot

src-tauri/src/profile/server/deploy.rs
```diff
@@ -0,0 +1,1308 @@
...
+use walkdir::{DirEntry, WalkDir};
+
+use super::{
+    config::RemoteServerSettings,
+    manifest::{self, DeploymentManifest, ManifestEntry},
+    remote::{self, ConnectionAttempt},
+};
+
+const MAX_MANIFEST_BYTES: u64 = 8 * 1024 * 1024;
+const MIRROR_DIRECTORIES: [&str; 3] = ["BepInEx/plugins", "BepInEx/patchers", "BepInEx/config"];
```
Kesomannen, line 22: This needs to be done in a game and mod-loader agnostic way.

Also, it's currently quite unclear when the mirror directories are being used. At first glance it looks like these are always the only directories considered. However, if I'm understanding the rest of the code correctly, these are _only_ used if BepInEx is already detected on the host. Please at least add comments explaining this, and maybe find some better naming (although for loader flexibility it can't be a constant).

src-tauri/src/profile/server/deploy.rs
```diff
@@ -0,0 +1,1308 @@
...
+}
+
+#[derive(Clone, Copy)]
+struct RemoteLayout {
+    bepinex_at_root: bool,
+}
+
+impl RemoteLayout {
+    fn path(self, base: &str, relative: &str) -> String {
+        let relative = if self.bepinex_at_root {
```
Kesomannen, line 125: Again, not flexible for other mod loaders. You'll need some other way to express this.

src-tauri/src/profile/server/deploy.rs
```diff
@@ -0,0 +1,1308 @@
...
+            relative.strip_prefix("BepInEx/").unwrap_or(relative)
+        } else {
+            relative
+        };
+        remote_path(base, relative)
+    }
+
+    fn manifest_path(self, base: &str) -> String {
+        if self.bepinex_at_root {
+            self.path(base, &format!("BepInEx/config/{}", manifest::FILE_NAME))
```
Kesomannen, line 135: If `bepinex_at_root` is set to true, isn't `BepInEx/config` at `config` and not with the `BepInEx` subdir?

src-tauri/src/profile/server/deploy.rs
```diff
@@ -0,0 +1,1308 @@
...
+        ConnectionAttempt::HostKeyUntrusted { fingerprint } => {
+            return Ok(DeploymentPreviewResult::HostKeyUntrusted { fingerprint });
+        }
+        ConnectionAttempt::CertificateUntrusted => {
+            return Ok(DeploymentPreviewResult::CertificateUntrusted);
+        }
+    };
+    let (local, plan, _) =
+        prepare_deployment(&mut connection, profile_dir, settings, &client_only_mods)?;
+    drop(local);
```
Kesomannen, line 159: Why bind `local` to immediately drop it?

src-tauri/src/profile/server/deploy.rs
```diff
@@ -0,0 +1,1308 @@
...
+    })
+}
+
+fn prepare_deployment(
+    connection: &mut remote::RemoteConnection,
+    profile_dir: &Path,
+    settings: &RemoteServerSettings,
+    client_only_mods: &BTreeSet<String>,
+) -> Result<(LocalDeployment, DeploymentPlan, RemoteLayout)> {
+    connection
```
Kesomannen, line 292: Please add some spaces between the lines here 😅

src-tauri/src/profile/server/deploy.rs
```diff
@@ -0,0 +1,1308 @@
...
+        && connection
+            .directory_exists(&remote_path(&settings.server_directory, "plugins"))
+            .context("failed to check for a restricted BepInEx server layout")?;
+    let server_has_bepinex = standard_layout
+        || bepinex_at_root
+        || connection
+            .directory_exists(&remote_path(&settings.server_directory, "BepInEx/core"))
+            .context("failed to check for a server-managed BepInEx installation")?;
+    let layout = RemoteLayout { bepinex_at_root };
+    let filter = ServerFileFilter::load(profile_dir, client_only_mods, server_has_bepinex);
```
Kesomannen, line 313: Computing the layout and filter could be extracted to its own function

src-tauri/src/profile/server/deploy.rs
```diff
@@ -0,0 +1,1308 @@
...
+        upload_bytes = plan.upload_bytes,
+        "prepared remote deployment"
+    );
+    Ok((local, plan, layout))
+}
+
+fn collect_remote_mirror_files(
+    connection: &mut remote::RemoteConnection,
+    base: &str,
+    layout: RemoteLayout,
```
Kesomannen, line 348: `connection`, `base` (or rather `settings`) and `layout` are passed to enough functions here to warrant a struct to hold them. These can then be members on that type.

src-tauri/src/profile/server/deploy.rs
```diff
@@ -0,0 +1,1308 @@
...
+                && entry.name != ".."
+                && !entry.name.contains(['/', '\\']),
+            "remote directory contains an unsafe name: {}",
+            entry.name
+        );
+        let relative = join_relative(directory, &entry.name);
+        if entry.is_directory {
+            directories.insert(relative.clone());
+            collect_remote_files(connection, base, &relative, layout, files, directories)?;
+        } else if relative != format!("BepInEx/config/{}", manifest::FILE_NAME)
```
Kesomannen, line 390: This would likely already get changed to accomodate other mod loaders, but if `bepinex_at_root` is set the manifest will be in `config` and not `BepInEx/config`, which breaks this check.

src-tauri/src/profile/server/deploy.rs
```diff
@@ -0,0 +1,1308 @@
...
+        if !entry.file_type().is_file() {
+            continue;
+        }
+
+        let relative = entry
+            .path()
+            .strip_prefix(profile_dir)
+            .context("profile file escaped its root")?;
+        if is_disabled(relative) {
+            disabled.insert(relative_path(&enabled_path(relative))?);
```
Kesomannen, line 600: Why use relative_path() here when the path is already relative?

src-tauri/src/profile/server/deploy.rs
```diff
@@ -0,0 +1,1308 @@
...
+            continue;
+        }
+        let relative = relative_path(relative)?;
```
Kesomannen, line 603: Same here

src-tauri/src/profile/server/deploy.rs
```diff
@@ -0,0 +1,1308 @@
...
+
+    Ok(value)
+}
+
+fn hash_file(path: &Path) -> Result<String> {
+    let mut reader = BufReader::new(
+        File::open(path).with_context(|| format!("failed to open {}", path.display()))?,
+    );
+    let mut hasher = blake3::Hasher::new();
+    let mut buffer = [0_u8; HASH_BUFFER_SIZE];
```
Kesomannen, line 706: Can just use std::io::copy here since blake3::Hasher implements Write

src-tauri/src/profile/server/deploy.rs
```diff
@@ -0,0 +1,1308 @@
...
+        } else if relative != format!("BepInEx/config/{}", manifest::FILE_NAME)
+            && relative != format!("BepInEx/config/{}.tmp", manifest::FILE_NAME)
+        {
+            files.insert(relative);
+        }
+    }
+    Ok(())
+}
+
+fn find_legacy_old_files(
```
Kesomannen, line 399: Why is this needed? Why would any .old files find themselves on the server without having gone through Gale (and thus being logged in the manifest file, instead of needing to be discovered like this)?

src-tauri/src/profile/server/deploy.rs
```diff
@@ -0,0 +1,1308 @@
...
+fn is_mirrored(path: &str) -> bool {
+    MIRROR_DIRECTORIES
+        .iter()
+        .any(|directory| path == *directory || path.starts_with(&format!("{directory}/")))
+}
+
+impl ServerFileFilter {
+    fn load(
+        profile_dir: &Path,
+        client_only_mods: &BTreeSet<String>,
```
Kesomannen, line 532: Take this as already owned since it gets cloned anyway

src-tauri/src/profile/server/deploy.rs
```diff
@@ -0,0 +1,1308 @@
...
+
+struct LocalDeployment {
+    files: BTreeMap<String, LocalFile>,
+    disabled: BTreeSet<String>,
+}
+
+#[derive(Default)]
+struct ServerFileFilter {
+    client_only_mods: BTreeSet<String>,
+    tracked_files: BTreeSet<PathBuf>,
```
Kesomannen, line 99: Not clear what "tracked_files" does here, there are better names like `client_only_files` or similar. Also, I'd change this to a HashMap since it doesn't need ordering (this goes for many other uses in this file).

src-tauri/src/profile/server/deploy.rs
```diff
@@ -0,0 +1,1308 @@
...
+            }
+        }
+        if self.tracked_files.contains(relative) {
+            return true;
+        }
+        let parts = relative
+            .components()
+            .filter_map(|part| part.as_os_str().to_str())
+            .collect::<Vec<_>>();
+        parts.len() >= 3
```
Kesomannen, line 570: The way client side filtering works currently is quite brittle. 

First of all, it relies on the string format of the categories which isn't guaranteed to stay the same across communities and platforms. For example, PEAK on Thunderstore doesn't use the same names, granted it doesn't have dedicated server support but it's probably not the only outlier.

Also, this is very mod-loader specific. Even with just BepInEx, this doesn't even cover the whole mod, just /plugins and isn't compatible with custom subdirectories. I'm having trouble seeing it scale to some of the other mod loaders in a nice, general way. It is kind of a design flaw Gale doesn't have a full-proof way of listing which files were installed by which mods. You're best bet with the current architecture would be to link it up to the mod install trait implementations.

But even then, I'd argue the benefit of this doesn't outweight the complexity and risk false positives of trying to filter out some mods' files. How concerned are you of people using the same profile for both servers and clients? How many mods client-mods completely break when ran on the server-side?

src-tauri/src/profile/server/deploy.rs
```diff
@@ -0,0 +1,1308 @@
...
+
+    let comparable_path = enabled_path(entry.path());
+    let Some(name) = comparable_path.file_name().and_then(|name| name.to_str()) else {
+        return false;
+    };
+    if name.eq_ignore_ascii_case("LogOutput.log") {
+        return true;
+    }
+
+    let plugin_index = components.windows(2).position(|parts| {
```
Kesomannen, line 676: This is very BepInEx-specific and should be solved in a more flexible way, if any at all. These files are usually quite small, so is it really a huge deal to not upload them? Additionally, some mods do read the manifest.json file of others which would be broken by this. I'd like to see the manifest.json file kept in the upload. Others can be excluded via globs rather than manual matching, which goes for this whole method. Take some inspiration from the filtering happening towards the end of `export/mod.rs`.

src-tauri/src/profile/server/deploy.rs
```diff
@@ -0,0 +1,1308 @@
...
+                thread::sleep(RETRY_DELAY * attempt as u32);
+                *connection = reconnect(settings, password)?;
+                ensured_directories.clear();
+            }
+        }
+    }
+    Err(last_error.unwrap())
+}
+
+fn write_manifest_with_retry(
```
Kesomannen, line 795: This should reuse upload_file_with_retry rather than reimplementing similar logic

src-tauri/src/profile/server/deploy.rs
```diff
@@ -0,0 +1,1308 @@
...
+    settings: &RemoteServerSettings,
+    password: &str,
+    layout: RemoteLayout,
+    base: &str,
+    relative: &str,
+    file: &LocalFile,
+    ensured_directories: &mut BTreeSet<String>,
+) -> Result<()> {
+    let mut last_error = None;
+    for attempt in 1..=TRANSFER_ATTEMPTS {
```
Kesomannen, line 766: The for-loop will never reach the end due to the 
```rust
if attempt == TRANSFER_ATTEMPTS {
    break;
}
```
clause. You can instead change it to a plain `loop` and do the return instead of break when the max attempts have been reached. This also saves the last_error variable and an unwrap().

src-tauri/src/profile/server/deploy.rs
```diff
@@ -0,0 +1,1308 @@
...
+        .write_file(&temporary, &bytes)
+        .context("failed to write remote Gale manifest")?;
+
+    replace_remote_file(connection, &temporary, &target)
+        .context("failed to replace remote Gale manifest")?;
+
+    Ok(())
+}
+
+fn replace_remote_file(
```
Kesomannen, line 907: The logic in this function makes sense but is a bit hard to follow. Some comments would be appreciated.

src-tauri/src/profile/server/local.rs
```diff
@@ -0,0 +1,146 @@
...
+    game.copy_required_files(&server_dir)
+        .context("failed to prepare mod loader files for dedicated server")?;
+
+    let profile = game.active_profile();
+    let mut command = Command::new(&executable);
+
+    if matches!(server_platform, Platform::Steam)
+        && let Some(steam) = &game.game.platforms.steam
+    {
+        command.env("SteamAppId", steam.id.to_string());
```
Kesomannen, line 49: What does this do/why is it required?

src-tauri/src/profile/server/local.rs
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    command
+        .current_dir(&server_dir)
+        .arg("-nographics")
+        .arg("-batchmode")
```
Kesomannen, line 55: Apart from the configurable Valheim-specific settings below that I've mentioned earlier, these also seem specific to Valheim.

src-tauri/src/profile/server/local.rs
```diff
@@ -0,0 +1,146 @@
...
+
+    game.apply_mod_loader_args(
+        &mut command,
+        &server_dir,
+        Some(server_platform),
+        &dedicated.platforms,
+    )
+    .context("failed to configure mod loader")?;
+
+    if !settings.extra_args.trim().is_empty() {
```
Kesomannen, line 81: Whitespace shouldn't matter for this method. Even then, I'd rather have it error on cases where it's only whitespace, rather than ignore the value silently. Discouraging/preventing such inputs from the user should be the frontend's responsibility.

src-tauri/src/profile/server/local.rs
```diff
@@ -0,0 +1,146 @@
...
+            .context("failed to apply dedicated server launch arguments")?;
+    }
+
+    #[cfg(target_os = "windows")]
+    {
+        use std::os::windows::process::CommandExt;
+
+        const CREATE_NEW_CONSOLE: u32 = 0x0000_0010;
+
+        command.creation_flags(CREATE_NEW_CONSOLE);
```
Kesomannen, line 92: Is this Valheim-specific?

src-tauri/src/profile/server/remote.rs
```diff
@@ -0,0 +1,571 @@
...
+                .map(|file| RemoteEntry {
+                    name: file.name().to_owned(),
+                    is_directory: file.is_directory(),
+                })
+                .with_context(|| format!("failed to parse FTP LIST entry: {entry}"))
+        })
+        .collect()
+}
+
+#[cfg(test)]
```
Kesomannen, line 450: Tests should be at the bottom of the file

src-tauri/src/profile/server/remote.rs
```diff
@@ -0,0 +1,571 @@
...
+    ClientConfig, DigitallySignedStruct, Error as TlsError, RootCertStore, SignatureScheme,
+    client::danger::{HandshakeSignatureValid, ServerCertVerified, ServerCertVerifier},
+    pki_types::{CertificateDer, ServerName, UnixTime},
+};
+use suppaftp::{FtpError, RustlsConnector, RustlsFtpStream, Status};
+
+use super::config::{RemoteAuthentication, RemoteProtocol, RemoteServerSettings};
+
+const CONNECT_TIMEOUT: Duration = Duration::from_secs(10);
+const SSH_TIMEOUT_MS: u32 = 15_000;
```
Kesomannen, line 23: Use a Duration here and convert it to milliseconds only at the call site

Kesomannen:
Still working on the review...

src-tauri/src/profile/server/deploy.rs
```diff
@@ -0,0 +1,1308 @@
...
+    }
+
+    fn excludes(&self, relative: &Path) -> bool {
+        if relative.as_os_str().is_empty() {
+            return false;
+        }
+        if self.preserve_server_bepinex {
+            let relative = relative.to_string_lossy().replace('\\', "/");
+            if !is_mirrored(&relative)
+                && !MIRROR_DIRECTORIES
```
Kesomannen, line 556: Isn't this doing the same thing as is_mirrored again?

src-tauri/src/profile/server/remote.rs
```diff
@@ -0,0 +1,571 @@
...
+    settings: &RemoteServerSettings,
+    password: &str,
+) -> Result<ConnectionAttempt> {
+    settings.validate()?;
+
+    if settings.protocol != RemoteProtocol::Sftp {
+        return match connect_ftp(settings, password) {
+            Ok(connection) => Ok(ConnectionAttempt::Connected(connection)),
+            Err(error)
+                if settings.protocol != RemoteProtocol::Sftp
```
Kesomannen, line 139: This is already checked by this point.

src-tauri/src/profile/server/remote.rs
```diff
@@ -0,0 +1,571 @@
...
+        }
+        ConnectionAttempt::CertificateUntrusted => Ok(ConnectionTestResult::CertificateUntrusted),
+    }
+}
+
+pub(crate) fn connect(
+    settings: &RemoteServerSettings,
+    password: &str,
+) -> Result<ConnectionAttempt> {
+    settings.validate()?;
```
Kesomannen, line 133: This goes hand in hand with the other settings changes I proposed, but we should validate once when this is set, instead of at every usage

src-tauri/src/profile/server/remote.rs
```diff
@@ -0,0 +1,571 @@
...
+        client: RemoteClient::Sftp {
+            sftp,
+            _session: session,
+        },
+        fingerprint: Some(fingerprint),
+        encrypted: true,
+    }))
+}
+
+fn connect_ftp(settings: &RemoteServerSettings, password: &str) -> Result<RemoteConnection> {
```
Kesomannen, line 179: Some spaces in this function would be appreciated :)

This could also be a function the RemoteConnection struct

src-tauri/src/profile/server/remote.rs
```diff
@@ -0,0 +1,571 @@
...
+                Err(err) if is_ftp_not_found(&err) => Ok(Vec::new()),
+                Err(err) => Err(err.into()),
+            },
+        }
+    }
+
+    pub fn check_directory(&mut self, path: &str) -> Result<()> {
+        match &mut self.client {
+            RemoteClient::Sftp { sftp, .. } => {
+                sftp.stat(Path::new(path)).map(|_| ()).map_err(Into::into)
```
Kesomannen, line 261: In my opinion, this is cleaner:

```suggestion
sftp.stat(Path::new(path))?;
Ok(())
```

src-tauri/src/profile/server/remote.rs
```diff
@@ -0,0 +1,571 @@
...
+fn host_key_fingerprint(session: &Session) -> Result<String> {
+    let hash = session
+        .host_key_hash(HashType::Sha256)
+        .ok_or_eyre("server did not provide a SHA256 host-key fingerprint")?;
+
+    Ok(format!("SHA256:{}", STANDARD_NO_PAD.encode(hash)))
+}
+
+fn authenticate(
+    session: &Session,
```
Kesomannen, line 540: Maybe qualify all ssh2 Sessions fully so it's harder to mix up with ftp?

src/lib/components/dialogs/DedicatedServerDialog.svelte
```diff
@@ -0,0 +1,558 @@
...
+	let trustedHostKey = $state<string | null>(null);
+	let trustedInvalidCertificateHost = $state<string | null>(null);
+	let remotePassword = $state('');
+	let rememberRemotePassword = $state(true);
+	let initialized = $state(false);
+	let loadingSettings = $state(false);
+	let saving = $state(false);
+	let launching = $state(false);
+	let testing = $state(false);
+	let deploying = $state(false);
```
Kesomannen, line 66: What do you think about collapsing these bool flags into one enum instead?

src/lib/components/dialogs/DedicatedServerDialog.svelte
Kesomannen, line 1: Do consider splitting this up to multiple components. For example the two pages (local and remote) could be split.

src/lib/components/dialogs/DeploymentPreviewDialog.svelte
```diff
@@ -0,0 +1,108 @@
...
+
+		<DeploymentStats
+			uploaded={preview.uploadFiles.length}
+			bytes={preview.uploadBytes}
+			removed={preview.removeFiles.length}
+			unchanged={preview.unchangedFiles}
+		/>
+
+		{#if preview.removeFiles.length > 0}
+			<InfoBox type="warning" class="mt-4">{m.deploymentPreview_removalInfo()}</InfoBox>
```
Kesomannen, line 45: This InfoBox is a bit weirdly placed. It's not clear it refers to the files within the `<details>` below. All in all I think the "Removals" section does a good enough job communicating what it does. I would also say it isn't that important to the user which exact files get removed or not (hence them being inside a `details` dropdown).

src-tauri/src/profile/server/secrets.rs
```diff
@@ -0,0 +1,58 @@
...
+            Self::GamePassword => "game-password",
+            Self::SftpPassword => "sftp-password",
+            Self::FtpPassword => "ftp-password",
+            Self::SshKeyPassphrase => "ssh-key-passphrase",
+        }
+    }
+}
+
+fn entry(profile_id: i64, secret: ServerSecret) -> Result<Entry> {
+    Entry::new(
```
Kesomannen, line 26: This creates a new entry everytime the secret is accessed. That _probably_ fine but I would like if they only needed to be initialized once.

src-tauri/src/profile/server/deploy.rs
```diff
@@ -0,0 +1,1308 @@
...
+        files: local
+            .files
+            .iter()
+            .map(|(path, file)| (path.clone(), file.manifest.clone()))
+            .collect(),
+    };
+    let uploads = local
+        .files
+        .iter()
+        .filter(|(path, file)| {
```
Kesomannen, line 460: This filter leads to some odd behaviour. On first sync, it doesn't do the mirror check (since previous.files.get(*path) returns None) which leads to all non-excluded files being uploaded. However, on next sync, they don't pass the first check, and since they're not mirrored, they get excluded and promptly deleted from the server. This happens to the BepInEx pack's core files.

messages/en.json
```diff
@@ -483,6 +590,7 @@
 	"toolBar_dialog_launch_vanilla_title": "Launching vanilla {name}...",
 	"toolBar_launch_modded": "Launch modded",
 	"toolBar_launch_vanilla": "Launch vanilla",
+	"toolBar_launch_server": "Launch dedicated server",
```
Kesomannen, line 593: "Launch server" is in my opinion better here as it's more concise and horizontal space is a bit constrained.

Kesomannen (changes requested):
Took a bit 😅 unfortunately there are quite a few things I would like you to check out, mostly related to code quality and game+mod loader flexibility. Apart from all the comments, there are also some bigger areas that cover a larger scope:

**Paths**

In much of the logic in `deploy.rs` and elsewhere, paths are handled as strings and not Paths. This leads to a lot of unneeded logic and much higher risk of subtle bugs or cross-platform issues. For example, checking if a path starts with a certain directory is much simpler and waterproof with a standard Path. This also extends to `RemoteServerSettings` which stores some path fields as strings.

Where possible, replace path variables and logic with the Path(Buf) types and only convert at the closest boundary (I'm guessing this will happen when sending the paths to ftp).

If using standard Paths gets too cumbersome due to string (utf8) conversions, feel free to add the `camino` crate, which has the specialized `Utf8Path(Buf)` for handling paths with only utf8.

**UX flow**

I think the current UX flow could be refined a bit. It's weird to have the settings dialog open every time you press Launch dedicated server. This isn't consistent with the other options which launch options which do the action immediately. Tweaking those settings is also something you won't do often, far from every launch.

My suggestion is to move the remote and local settings to a dialog that you access from the profile settings (at the bottom of the settings page). Semantically it makes sense to put near the platform/launch method/custom arguments. That is except for the first time you press the launch server option, where I think it's fine to open the dialog for the initial setup.

**Sync remote connections**

The ftp and ssh libraries are currently sync while the rest of the app is async. This is likely fine, but could lead to pretty catastrophic blocking if a `tokio::task::spawn_blocking` is ever missed. It would be nice to investigate async alternatives. However if switching to async is too much of a burden, I wouldn't bother and go with the current approach.

Also, make sure to run rustfmt, there are a couple of spots that look unformatted.

Hope you can find the time to go through and look at all the comments. If you want to hold a discussion a little faster, we can take it in the hexium discord rather than here on github. I also apologize for the huge number of notes, but this is a big feature so I want to get it right :)

AzumattDev:
Thanks for the detailed review. The feature grew out of testing against Valhost and G-Portal, where restricted directory layouts, FTP limitations, TLS behavior, and partially managed BepInEx installations drove quite a few implementation decisions.
I generalized platform discovery and the transport/deployment pipeline, but the review correctly identifies that the local launch arguments and deployment policy are still too Valheim/BepInEx-specific. Since I'm most familiar with that game and the setup, it only made sense to structure it that way as I'm not fully aware of how other games handle it. A few games I have seen seem to handle things very similarly. I also agree that the commands and primary dialog accumulated too many responsibilities as the implementation expanded. It made sense for me in testing as I was testing with multiple hosts at the same time...but I see now that it doesn't outside of that.

My priorities will be:
1. Fix the deployment ownership bug that can upload BepInEx core files on one deployment and remove them on the next.
2. Separate generic server process/transport handling from game- and mod-loader-specific launch/deployment policy.
3. Move persistent setup into profile settings and make an already-configured launch immediate.
4. Tighten path handling and isolate synchronous network work behind one blocking boundary.
5. Apply the smaller codebase-consistency and UI comments.
A few of the restricted-root path observations are already handled by the logical-path-to-remote-path mapping, but the current naming makes that unnecessarily difficult to see. I’ll clarify those cases below and make the representation clearer in code.

AzumattDev:
> Took a bit 😅 unfortunately there are quite a few things I would like you to check out, mostly related to code quality and game+mod loader flexibility. Apart from all the comments, there are also some bigger areas that cover a larger scope:
> 
> **Paths**
> 
> In much of the logic in `deploy.rs` and elsewhere, paths are handled as strings and not Paths. This leads to a lot of unneeded logic and much higher risk of subtle bugs or cross-platform issues. For example, checking if a path starts with a certain directory is much simpler and waterproof with a standard Path. This also extends to `RemoteServerSettings` which stores some path fields as strings.
> 
> Where possible, replace path variables and logic with the Path(Buf) types and only convert at the closest boundary (I'm guessing this will happen when sending the paths to ftp).
> 
> If using standard Paths gets too cumbersome due to string (utf8) conversions, feel free to add the `camino` crate, which has the specialized `Utf8Path(Buf)` for handling paths with only utf8.
> 
> **UX flow**
> 
> I think the current UX flow could be refined a bit. It's weird to have the settings dialog open every time you press Launch dedicated server. This isn't consistent with the other options which launch options which do the action immediately. Tweaking those settings is also something you won't do often, far from every launch.
> 
> My suggestion is to move the remote and local settings to a dialog that you access from the profile settings (at the bottom of the settings page). Semantically it makes sense to put near the platform/launch method/custom arguments. That is except for the first time you press the launch server option, where I think it's fine to open the dialog for the initial setup.
> 
> **Sync remote connections**
> 
> The ftp and ssh libraries are currently sync while the rest of the app is async. This is likely fine, but could lead to pretty catastrophic blocking if a `tokio::task::spawn_blocking` is ever missed. It would be nice to investigate async alternatives. However if switching to async is too much of a burden, I wouldn't bother and go with the current approach.
> 
> Also, make sure to run rustfmt, there are a couple of spots that look unformatted.
> 
> Hope you can find the time to go through and look at all the comments. If you want to hold a discussion a little faster, we can take it in the hexium discord rather than here on github. I also apologize for the huge number of notes, but this is a big feature so I want to get it right :)


# My response to these, in order/respectfully.

Paths
> Local filesystem values should absolutely remain Path/PathBuf.
Remote paths were represented as strings deliberately because they belong to a remote POSIX-style namespace even when Gale is running on Windows. Using PathBuf::join on Windows can introduce host-native separators into paths sent to FTP/SFTP.
That said, the current implementation lets string manipulation spread much too far. I’ll introduce one canonical remote/deployment path representation, enforce forward-slash relative paths there, and convert to the FTP/SFTP library’s required type only at the transport boundary. Manifest keys will use that same canonical representation.

UX flow
> The dialog originally served three purposes: first-time setup, trust confirmation/connection testing, and deployment preview. Keeping those together made the safety flow obvious during testing, but it also made the normal local-launch path unnecessarily modal.
I agree with the proposed flow and that persistent settings belong under profile settings. The first launch can open setup and subsequent local launches should run immediately using whatever we have in profile settings. Remote deployment can still show its preview because that is an operation-specific confirmation rather than routine configuration. I also think we should at bare minimum allow quick changes to the configuration without leaving the modal, for pure QOL on the user.

Synchronous FTP/SFTP
> The synchronous libraries were intentional. ssh2 provided the required host-key handling, while the selected FTP stack supported explicit FTPS, self-signed-certificate confirmation, and the data-channel TLS behavior encountered on real hosting providers.
All remote test, preview, and deployment entry points currently run under spawn_blocking, so the Tokio executor is not blocked. The real maintainability issue is that every caller must remember that requirement. I’ll centralize remote work behind one blocking adapter and evaluate async replacements against the same SFTP/FTPS compatibility requirements before changing libraries.

src-tauri/src/profile/launch/mod.rs
AzumattDev: I’ll use bare pub to match the crate’s convention.

src-tauri/src/profile/launch/mod.rs
AzumattDev: I see...I'll change.

src-tauri/src/profile/server/runtime.rs
AzumattDev: I addressed this in my bigger comment, but it shouldn't be blocking

dannzen:
stupid question, how are you stopping or starting the g-portal server?

AzumattDev:
> stupid question, how are you stopping or starting the g-portal server?

Not a stupid question I'm currently not. Gale only starts and stops dedicated servers launched locally. For G‑Portal, remote support is profile deployment over FTP/FTPS. Users must stop the server through G‑Portal before deploying and start it again afterward. G‑Portal does not appear to expose a documented control API that can safely integrate.

Kesomannen:
> That said, the current implementation lets string manipulation spread much too far. I’ll introduce one canonical remote/deployment path representation, enforce forward-slash relative paths there, and convert to the FTP/SFTP library’s required type only at the transport boundary. Manifest keys will use that same canonical representation.

Using Paths basically everywhere and just converting from `\` to `/` (as well as checking for UTF-8 compatibility) at the ftp boundary looks to be enough. The ssh2 methods already take `AsRef<Path>`. I ran a quick test on Windows to test if it actually accepted backslashes with a linux ssh server and the following does work:

```rust
use std::{net::TcpStream, path::Path};

use ssh2::Session;

fn main() -> Result<(), Box<dyn std::error::Error>> {
    let tcp = TcpStream::connect("url:22")?;

    let mut sess = Session::new()?;
    sess.set_tcp_stream(tcp);
    sess.handshake()?;
    sess.userauth_password("user", "...")?;

    let sftp = sess.sftp()?;
    let entries = sftp.readdir(Path::new(r"\home\user"))?;
    println!(
        "{:#?}",
        entries
            .iter()
            .map(|entry| entry.0.display())
            .collect::<Vec<_>>()
    );

    Ok(())
}
```

> I also think we should at bare minimum allow quick changes to the configuration without leaving the modal

What do you mean by this? That you can configure some settings from the preview dialog that always opens?

> All remote test, preview, and deployment entry points currently run under spawn_blocking, so the Tokio executor is not blocked. The real maintainability issue is that every caller must remember that requirement. I’ll centralize remote work behind one blocking adapter and evaluate async replacements against the same SFTP/FTPS compatibility requirements before changing libraries.

:+1:

Issue #726, opened 2026-09-11
Gale is not upgrading thunderstore mods when update exists on hexium.


hardtimes:
AzuEPI is stuck at 2.4.8 when download source in gale is set to hexium + thunderstore. Only way to get it to update is to remove and install 2.4.13 from hexium. 

As reported in the discord:
Undivided — 11:55 AM
Double checked with a new profile with only that mod. Installed with only thunderstore enabled. enabled both thunder and hexium. exited and re-entered gale. this is the result(no update available for azuapi 2.4.8). Double checked thunder and hexium are still the download sources.

<img width="1366" height="1171" alt="Image" src="https://github.com/user-attachments/assets/cfd68ac4-e7b2-4110-9993-9e44e9fc7dfe" />

jterrace:
Coming here from Discord - I'm not so sure the version comparison is the only problem. Two mods I'm seeing right now that aren't updating that would have compared fine with string comparisons:

 - https://thunderstore.io/c/valheim/p/Azumatt/Where_You_At/ - 1.0.11 on Thunderstore
 - https://valheim.hexium.gg/mods/Azumatt/Where_You_At - 1.0.12 on Hexium
 - https://thunderstore.io/c/valheim/p/RustyMods/MarketStall/ - 1.1.4 on Thunderstore
 - https://valheim.hexium.gg/mods/RustyMods/MarketStall - 2.0.0 on Hexium

Despite having Thunderstore + Hexium set as mod source, I'm not seeing updates on these two mods. They're stuck on the Thunderstore versions.

Kesomannen:
This should be fixed in 1.22.3, let me know if you have any continued issues or not

jterrace:
Hmm, after updating to 1.22.3, Where_You_At is now showing an update from 1.0.11 to 1.0.12. However, MarketStall is still showing no update with 1.1.4 installed. Could it be because it's marked as deprecated on Thunderstore?

Kesomannen:
> Hmm, after updating to 1.22.3, Where_You_At is now showing an update from 1.0.11 to 1.0.12. However, MarketStall is still showing no update with 1.1.4 installed. Could it be because it's marked as deprecated on Thunderstore?

I'll investigate

Kesomannen:
I see, Gale doesn't consider cross-platform updates on the individual mod cards, and only adds the update to the "Update All" banner:

[...]

PR #734, opened 2026-09-12
Fix string version comparison

bwoebi:
This caused e.g. 2.4.8 > 2.4.14.


Kesomannen:
Ah right, nice catch. That should fix https://github.com/Kesomannen/gale/issues/726

bwoebi:
Indeed, I did not realize there was a separate bug report for it already.

PR #736, opened 2026-09-13
fix: remove empty package directories when clearing unused cache

### LINKED ISSUE #735: Not removing empty cache directories
When the last version of a mod is cleared from the cache with "Clear unused mod cache" the empty mod directory is left in the cache directory. While this is a minor issue only likely to cause problems in extremely rare cases, it would still be nice to remove the now empty directory.

Tested on Gale 1.22.2 AppImage for Linux.

jamalkamaladdin:
do_soft_clear removes each cached version directory, and the package directory above it stays behind once its last version is gone.

The package directory is now removed as well when it is left empty. A failure to remove it is logged and does not stop the clear.

Closes #735


src-tauri/src/profile/install/cache.rs
```diff
@@ -80,11 +80,29 @@ pub(super) fn prepare_soft_clear(app: AppHandle) -> Result<Vec<PathBuf>> {
...
 pub(super) fn do_soft_clear(paths: Vec<PathBuf>) -> Result<()> {
     let count = paths.len();
 
     for path in paths {
+        let package_dir = path.parent().map(Path::to_path_buf);
+
         fs::remove_dir_all(path)?;
+
+        // Clearing the last version of a mod leaves its package directory behind
+        let Some(dir) = package_dir else { continue };
```
Kesomannen: These brackets shouldn't be on the same line. Make sure to run `cargo fmt`

src-tauri/src/profile/install/cache.rs
```diff
@@ -80,11 +80,29 @@ pub(super) fn prepare_soft_clear(app: AppHandle) -> Result<Vec<PathBuf>> {
...
+fn is_empty_dir(path: &Path) -> bool {
+    path.read_dir()
+        .is_ok_and(|mut entries| entries.next().is_none())
+}
+
 pub(super) fn do_soft_clear(paths: Vec<PathBuf>) -> Result<()> {
     let count = paths.len();
 
     for path in paths {
+        let package_dir = path.parent().map(Path::to_path_buf);
```
Kesomannen: If you change the line below to `&path` the clone here isn't required.

Kesomannen:
:+1:

PR #752, opened 2026-09-17
Show whether the upload was hidden

bwoebi:
This hidden: bool on successful upload is exclusive to Hexium, hence Option<bool>.

We do hide new uploads by default, to prevent a leaked API key from being used to upload new versions and make them available to everyone without the uploader explicitly making the uploaded version visible on Hexium.

Users can disable this in their settings, but we prefer a secure default. It happens though often enough that people forget to unhide them on Hexium. This serves as a reminder.


src-tauri/src/thunderstore/models.rs
```diff
@@ -219,6 +219,11 @@ pub struct UserMediaFinishUploadParams {
     pub parts: Vec<CompletedPart>,
 }
 
+#[derive(Deserialize, Serialize, Debug)]
+pub struct UploadSubmissionResult {
+    pub hidden: Option<bool>,
```
Kesomannen, line 224: Pretty sure you need #[serde(default)] here, otherwise it will fail on Thunderstore

bwoebi: Actually not: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7ecdefd0f0b00f70d969571b36a5f50e (taken from somewhere else) - serde_json assumes missing to be Option::None by default.

Shall I still change it (like: do you prefer it that way)?

Kesomannen: Oh ok, you learn something new every day. Then it's fine to keep as this. And to double check, you are sure thunderstore returns json at all?

bwoebi: I did not exercise the endpoint, but according to https://thunderstore.io/api/docs/ it returns a PackageSubmissionResult object (hexium API includes the same).

messages/en.json
```diff
@@ -322,6 +322,7 @@
 	"modpack_dialog_content_2": "Click here to view its page on the website.",
 	"modpack_dialog_content_3": "The changes may take up to an hour to appear in Gale and other mod managers.",
 	"modpack_dialog_content_4": "To publish a new update, increment the version number and publish the modpack again.",
+	"modpack_dialog_content_hidden": "Don't forget to unhide your mod pack on {backend}, so that your friends can search for it.",
```
Kesomannen: Currently it doesn't really explain why the mod is hidden or particularly how you fix it. How about:
```suggestion
By default, your modpack is hidden from other users. Visit the {backend} website to unhide it, so that your friends can search for it.
```

bwoebi: Woops, that replaced the whole line

bwoebi: Fixed.

Kesomannen: > Woops, that replaced the whole line

Ah, my mistake, I left the key and stuff out of the suggestion block

PR #758, opened 2026-09-19
feat(custom_launch_arguments): resolve multiple components for prefix

avnyu:
Allows something like this to work, which similar to Steam custom launch:
`env -u DISPLAY mangohud %command%`
The current workaround is to write all as single script and use it as prefix.

This PR:
Changed prefix from `Option<String>` to `Vec<String>`, if prefix is not empty, use the first as cmd and the rest as arguments directly after.
Added test for prefix_multi and apply_prefix_multi.

This also allow unset environment variable using `env -u`, as the example above.
I not extensively tested this yet, but work for my case, and it's already passed related tests.

Currently test `config::bepinex::tests::check_from_string` failed, but it seem unrelated to this PR and I also failed to run the test on HEAD.


src-tauri/src/profile/launch/custom_args.rs
```diff
@@ -101,15 +102,14 @@ impl FromStr for CustomArgs {
 
         let mut args = Vec::new();
         let mut env = Vec::new();
-        let mut prefix = None;
+        let mut prefix = Vec::new();
```
jamalkamaladdin: `prefix.append(&mut args)` leaves `prefix` empty when the first `%command%` has nothing before it, and the next `%command%` then passes `!prefix.is_empty()`.

Measured with rustc 1.98.1, this loop against the `Option<String>` loop on master:

`%command% --foo %command% --bar` gives `Err("multiple %command% placeholders are not allowed")` on master, `Ok(args=["--bar"], prefix=["--foo"])` here

`%command% %command%` is accepted here

`multiple_prefixes_fail` does not cover it: its input starts with `protontricks`, which fills `prefix`.

```suggestion
        let mut prefix = Vec::new();
        if words.iter().filter(|w| *w == "%command%").count() > 1 {
            bail!("multiple %command% placeholders are not allowed");
        }
```

jamalkamaladdin (commented):
One measured note below.

src-tauri/src/profile/launch/custom_args.rs
avnyu: Thanks for review, I kinda want to rewrote the for logic part in `from_str`, but decided not to do so, and forget about this afterward. If the `Vec` empty state could mean both no prefix and empty prefix, I guess I should change change to `Option<Vec<String>>` instead? `None` for no prefix, `Some(vec)` for existed prefix, whether it empty or not? Or I should follow this suggestion also add the test for this.

jamalkamaladdin: `Option<Vec<String>>` works. ran your loop against the cases from my first comment:

  `%command% --foo %command% --bar` and `%command% %command%` both bail now. `%command% --foo` gives `prefix=Some([])` and `apply` skips it, which also kills master's `Command::new("")`.

  One case your tests miss. A prefix word containing `=` never reaches `prefix`:

  `mangohud --dlsym=1 %command%` gives `prefix=["mangohud"] env=[("--dlsym", "1")]`

  `env -u DISPLAY mangohud --fps-limit=60 %command%` gives `prefix=["env", "-u", "DISPLAY", "mangohud"] env=[("--fps-limit", "60")]`

  `split_once('=')` grabs the word before position matters. master had the same branch, but master could not run a multi word prefix anyway.

avnyu: That's hard. I don't think add check for `-` at start position is enough.

If the expect is everyone put all `X=Y` at the start, then I can extract the prefix `X=Y` to add to `env`, but as pointed, position do not matters at master.

For clean solution, maybe moving environment variables into new `Custom environment variables` field? But then it's a breaking changes.

jamalkamaladdin: No need to move env out or care about position. Env var names can only be `[A-Za-z_][A-Za-z0-9_]*`. A word like `--dlsym=1` was never a valid env var. Check the key before treating it as
  env:

  ```rust
  fn is_env_name(s: &str) -> bool {
      let mut chars = s.chars();
      matches!(chars.next(), Some(c) if c == '_' || c.is_ascii_alphabetic())
          && chars.all(|c| c == '_' || c.is_ascii_alphanumeric())
  }
  ```

  ```rust
  } else if let Some((key, value)) = word.split_once('=').filter(|(k, _)| is_env_name(k)) {
  ```

  Ran your loop with that change:

  `mangohud --dlsym=1 %command%` gives `prefix=["mangohud", "--dlsym=1"] env=[]`
  `FOO=bar mangohud --dlsym=1 %command% --x BAZ=qux` gives `prefix=["mangohud", "--dlsym=1"] args=["--x"] env=[FOO, BAZ]`
  `mangohud FOO=1 %command%` still puts `FOO` in env, same as master
  `--foo BAR=baz` without a prefix is unchanged

  Only other change: a bare `--opt=val` now reaches the game as an arg. Master turned it into an env var named `--opt`.

avnyu: I mean some command may have the value for argument in the same form as env var, for example `cmd -o option1=value1 -o option2=value2` and if not care about position, `option1=value1` get treated as env var? But I currently can not think of any examples.
I will force-push your suggest first.

src-tauri/src/profile/launch/custom_args.rs
```diff
@@ -108,9 +109,9 @@ impl FromStr for CustomArgs {
                 if prefix.is_some() {
                     bail!("multiple %command% placeholders are not allowed");
                 }
-                prefix = Some(args.join(" "));
-                args.clear();
-            } else if let Some((key, value)) = word.split_once('=') {
+                prefix = Some(args);
+                args = Vec::new();
+            } else if let Some((key, value)) = word.split_once('=').filter(|(k, _)| is_env_name(k)) {
```
Kesomannen: I think `is_some_and` is clearer here

Kesomannen (commented):
Looks reasonable, just one tiny nitpick

src-tauri/src/profile/launch/custom_args.rs
avnyu: Do you mean to replace `filter` by `is_some_and`?
`is_some_and` return `bool`, but we need `(key, value)` to be pushed to `env` at the line below, hence using `if let Some((key, value))` and `filter` instead?

